PR #3, opened 2011-04-11
Fix a failing test.

agnellvj:
the Search#to_curl method didn't join the indices to convert them to a comma delimited string which caused a test to fail.


karmi:
Thanks!, added test and merged in cae1f5505109109ec2dd3679470914dbc1c445b2.

PR #17, opened 2011-05-09
Properties _type and _index are nil

vhyza:
Collection items have _index and _type methods, but they are nil. This commit should resolve this.


karmi:
merged, thanks!

PR #18, opened 2011-05-09
Tire sometimes returns wrong total_pages value

vhyza:
total_pages should be the smallest integer greater of equal than total_entries / per_page
example: 4.0 / 3.0 = 1.333 -> total pages should be 2


karmi:
merged, thanks!

PR #23, opened 2011-05-29
Missing testing dependency

filiptepper:
Hi,

Just a little fix for a missing test dependency and Ruby 1.9 compatibility.

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 scratch/
 examples/*.html
 *.log
+.rvmrc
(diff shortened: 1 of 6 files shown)


karmi:
Hi, thanks! I've cleaned up and merged the 1.9.2 related commits, see https://github.com/karmi/tire/commits/master.

I didn't update the Rake and Rails dependencies, since I don't think the gem could depend on prerelease versions, and, more importantly, I would like to redo the dependencies anyway. Probably removing the Rake dependency altogether, and removing the ActiveModel dependency with a warning to the user when ActiveModel is not already loaded.

PR #24, opened 2011-05-30
default_operator

seejohnrun:
Added support to #string for :default_operator


karmi:
Hi, thanks, squashed and merged in karmi/tire@6a3b4f3. However, it was later superseded by @vhyza's patch at karmi/tire@7d57038 :)

seejohnrun:
Cool - I was going to make that move but was assuming you had excluded the rest of the operators for a reason.
Thanks for the gem, keep up the great work

karmi:
No, I just try to stick "implement only what you need" philosophy. In almost any case when I diverged from that it was pain :)

seejohnrun:
Yeah its definitely better for the gem too, especially as ES expands their API

PR #25, opened 2011-05-30
Missing _type and _index attributes

vhyza:
When finding documents by id, _index and _type attributes are missing


karmi:
Thanks, merged in karmi/tire@fa93077a5baebc3bb3d3

PR #26, opened 2011-05-31
Allow to set search options when using a query string

vhyza:
Now its possible to set default_field option only. This commit should allow more options like fields, use_dis_max etc (http://www.elasticsearch.org/guide/reference/query-dsl/query-string-query.html)


karmi:
Thanks!, merged in karmi/tire@7d57038

PR #27, opened 2011-06-01
Added open/close index functionality.

bcoe:
Added _open/_close index functionality. This is good for multitenant applications where not all lucene indexes are open at once.

I need this for my production deployment of elasticsearch, would love to use the canonical project, and not my fork :)

### DIFF
--- a/examples/dsl.rb
+++ b/examples/dsl.rb
@@ -51,7 +51,7 @@ puts "", "Query:", "-"*80
 puts s.to_json
 
 puts "", "Raw JSON result:", "-"*80
-puts JSON.pretty_generate(s.response)
+puts s.response
 
 puts "", "Try the query in Curl:", "-"*80
 puts s.to_curl
--- a/lib/tire.rb
+++ b/lib/tire.rb
@@ -1,7 +1,6 @@
 require 'rest_client'
-require 'yajl/json_gem'
 require 'active_model'
-
+require 'tire/rubyext/json'
 require 'tire/rubyext/hash'
 require 'tire/rubyext/symbol'
 require 'tire/logger'
--- a/lib/tire/dsl.rb
+++ b/lib/tire/dsl.rb
@@ -16,7 +16,7 @@ module Tire
         end
 
         response = Configuration.client.post( "#{Configuration.url}/#{indices}/_search", payload)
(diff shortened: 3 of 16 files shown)


bcoe:
The yajl-ruby library was causing major issues in my Rails deployment. The main problem being that the :methods key appears to not be supported in their JSON extension. I modified Tire so that it defaults to the 'json' gem by default, but can still use yajl if available (and working). I also found one other bug in my production deployment related to arrays having a 'to_hash' method punched onto them, I submitted a patch for this -- Tire is working great for me now in production.

karmi:
Hi, thanks! I've squashed the commits into into one, and edited. Merged in karmi/tire@d51f0cc1d143eb7907a1c30cd581335ad4babee0.

I am aware that the duplication in the execute > rescue > raise > ensure is quite dumb. However, I don't like having the wrapper method like you did there. There's certainly a better way, most probably in augmenting the `Client` class, which is only a thin wrapper for now.

karmi:
Regarding the other two issues:

1/ Yes, forcing yajl on everybody is absolutely stupid. Rails now goes with the intridea/multi_json gem, so I'll follow suit.

2/ Wow, Array having a `to_hash` method? This is coming from some library? @floere corrected the previous check for `is_a?(Hash)` (which is what you should have used?), see commit tire@081eb36. I think we can return to previous version. Could you find the library which implements `Array#to_hash`?

karmi:
OK, the yajl dependency removed in tire@90011cb4e1b3bb29666b75beb01ed50213029e78.

bcoe:
Cool, yajl was causing issues for me :) 

I'm not quite sure what is punching to_hash onto an Array. But, I think relying on that method name to determine if something is a hash is a bit cavalier. It's just too loaded, i.e., it might refer to just creating an MD5 hash out of a non-hashmap object.

Thanks for the updates :)

karmi:
>  relying on that method name to determine if something is a hash is a bit cavalier

LOL, definitely! I'll change it to the previous version, no harm by that and you'll be sorted.

bcoe:
Thanks :) very much appreciated,

karmi:
OK, all issues should be fixed and pushed at tire@v0.1.5, also released to Rubygems. Cheers!

PR #28, opened 2011-06-05
[Bug Fix] to_json must accept optional generator

bcoe:
[Bug Fix] Some JSON libraries require an optional generator be passed into the to_json method. This makes the addition of multi_json solve a more general set of problems.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -189,8 +189,69 @@ count for articles tagged 'php' is excluded, since they don't match the current 
     # java       1
 ```
 
-If configuring the search payload with a block somehow feels too weak for you, you can simply pass
-a Ruby `Hash` (or JSON string) with the query declaration to the `search` method:
+Notice, that only variables from the enclosing scope are accesible.
+If we want to access the variables or methods from outer scope,
+we have to use a slight variation of the DSL, by passing the
+`search` and `query` objects around.
+
+```ruby
+    @query = 'title:T*'
+
+    Tire.search 'articles' do |search|
+      search.query do |query|
+        query.string @query
+      end
+    end
+```
+
+Quite often, we need complex queries with boolean logic.
+Instead of composing long query strings such as `tags:ruby OR tags:java AND NOT tags:python`,
+we can use the [_bool_](http://www.elasticsearch.org/guide/reference/query-dsl/bool-query.html)
+query. In _Tire_, we build them declaratively.
+
+```ruby
+    Tire.search('articles') do
+      query do
+        boolean do
+          should   { string 'tags:ruby' }
+          should   { string 'tags:java' }
+          must_not { string 'tags:python' }
+        end
+      end
+    end
+```
+
+The best thing about `boolean` queries is that we can easily save these partial queries as Ruby blocks,
+to mix and reuse them later. So, we may define a query for the _tags_ property:
+
+```ruby
+    tags_query = lambda do
+      boolean.should { string 'tags:ruby' }
+      boolean.should { string 'tags:java' }
+    end
+```
+
+And a query for the _published_on_ property:
+
+```ruby
+    published_on_query = lambda do
+      boolean.must   { string 'published_on:[2011-01-01 TO 2011-01-02]' }
+    end
+```
+
+Now, we can combine these queries for different searches:
+
+```ruby
+    Tire.search('articles') do
+      query do
+        boolean &tags_query
+        boolean &published_on_query
+      end
+    end
+```
+
+If configuring the search payload with blocks somehow feels too weak for you, you can pass
+a plain old Ruby `Hash` (or JSON string) with the query declaration to the `search` method:
 
 ```ruby
     Tire.search 'articles', :query => { :fuzzy => { :title => 'Sour' } }
--- a/Rakefile
+++ b/Rakefile
@@ -25,11 +25,8 @@ namespace :test do
 end
 
 # Generate documentation
-begin
-  require 'sdoc'
-rescue LoadError
-end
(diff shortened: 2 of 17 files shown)


karmi:
Yeah, this is ActiveSupport, most probably, where `to_json` takes a `Hash` of options. However, I don't get any error for this, since this options Hash is optional. Do you have a specific error, gem, etc., where it fails?

bcoe:
It fails for me in my production Rails 3 environment with an error like this: (1 argument provided 0 expected). This exception is raised in all the the Tire classes with the to_json method. Not quite sure which library is causing this exception, my guess would be it's activesupport (3.0.7) -- simply accepting a parameter and throwing it away allows me to run example/dsl.rb in production.

bcoe:
[REQUEST FAILED] ArgumentError 
ArgumentError: wrong number of arguments (1 for 0)
    from /usr/lib/ruby/gems/1.8/gems/tire-0.1.6/lib/tire/search.rb:94:in `to_json'
    from /usr/lib/ruby/gems/1.8/gems/tire-0.1.6/lib/tire/search.rb:94:in`to_json'
    from /usr/lib/ruby/gems/1.8/gems/tire-0.1.6/lib/tire/search.rb:81:in `to_curl'
    from /usr/lib/ruby/gems/1.8/gems/tire-0.1.6/lib/tire/search.rb:74:in`perform'
    from /usr/lib/ruby/gems/1.8/gems/tire-0.1.6/lib/tire/dsl.rb:10:in `search'
    from (irb):10
    from /usr/local/lib/site_ruby/1.8/rubygems/exceptions.rb:19

There you go :) accepting the parameters and throwing them out has been tested for me and fixes this issue.

karmi:
This is really weird, since we run a Rails 3.0.7 app in production as well, and the demo app (`rails new searchapp -m https://github.com/karmi/tire/raw/master/examples/rails-application-template.rb`) runs fine as well. I'll have another look into this...

bcoe:
Hmm, I think it's actually the json (1.5.1) gem that expects the generator -- try against this in irb. Again, accepting the optional parameter solves this incompatibility issue.

Any ways, time to go start work on my own JSON library -- that will solve the problem!

karmi:
I am still not convinced this is the case. Try to reproduce it like this.

First, generate a demo Rails 3.0.7 app:

```
rails new searchapp -m https://github.com/karmi/tire/raw/master/examples/rails-application-template.rb
```

Then, change the JSON library to the "json" gem in your `Gemfile`:

```
gem "json"
```

and load the JSON library in your `application.rb`:

```
require 'rails/all'
require 'json'
```

Now, does searching in the app work for you or not? Because in my case, I get no `ArgumentError`. And rightly so, because the "json" gem really does have any arguments for `to_json` as optional. Evidence: https://github.com/flori/json/blob/master/lib/json/pure/generator.rb#L257

So, _if_ it does fail for you in the example app, then we'll dig into this deeper. I'm rather reluctant to add all the throwaway arguments to all the `to_json` methods. I'm happy that _Tire_ does not depend on "yajl" now, but wouldn't like to over-compicate stuff.

bcoe:
Try the 'json' dependency without requiring 'rails/all'. The only dependencies I'm including in irb are:

require 'rubygems'
require 'json'
require 'tire'

karmi:
Yeah, I've found this as well on my own. Seems like requiring Rails, ie. requiring ActiveSupport somehow makes the good ole' "json" gem to behave... I think adding the throwaway args to `to_json` methods in _Tire_ is the only solution, really.

bcoe:
Lol, with regards to my blog post, I hope there's no hard feelings -- really wasn't trying to go after your library :) I'm going to keep using/patching it.

karmi:
No, no hard feelings whatsoever! I just felt I should reply there, since a friend referred me to your article. I think you were too harsh on Ruby and its ecosystem, so I replied. And the more I think about the more I wonder why `to_json` methods in Tire with arity==0 don't fail at all times :)

bcoe:
My branch has diverged fairly far from master, don't see a point to keeping this open.

znbailey:
I just spun up a new rails project using tire and I'm running into this issue, though the error is slightly different:

wrong number of arguments (2 for 1)

stack is:

activesupport (3.1.2) lib/active_support/core_ext/object/to_json.rb:19:in `to_json'
tire (0.3.11) lib/tire/index.rb:296:in`to_json'
tire (0.3.11) lib/tire/index.rb:296:in `convert_document_to_json'
tire (0.3.11) lib/tire/index.rb:54:in`store'
tire (0.3.11) lib/tire/index.rb:8:in `instance_eval'
tire (0.3.11) lib/tire/index.rb:8:in`initialize'

What's the recommended approach to getting everything playing nice together? This project literally has one controller file in it and that's it (no complicated dependencies). Running ruby 1.9.2 and the only gems in my gemfile are rails, tire, and sqlite3

Thanks!

znbailey:
Looks like adding a dependency on yajl-ruby in my Gemfile and then adding:

require 'yajl/json_gem'

before

require 'rails/all' 

in my application.rb fixed things up. Not entirely sure why as it seems like multi_json should be falling back to the standard json gem?

karmi:
Thanks for the report -- this issue is really hard to hunt down, since different components overload JSON serialization/behaviour differently. Could you try to to change the Hash#to_json definition in https://github.com/karmi/tire/blob/master/lib/tire/rubyext/hash.rb#L3 to:

``` ruby
def to_json(options = nil)
  MultiJson.encode(self)
end unless respond_to?(:to_json)
```

PR #42, opened 2011-06-21
Populate results collection lazily & fixed two TODOs

clemens:
Hey Karel,

I've noticed that results are iterated automatically when the results collection is instantiated. IMO this should be done lazily because this allows better performance if – for whichever reason – you're not interested in the results but just the meta data and/or facets (I have the latter use case in an app that's currently using Thinking Sphinx).

If you prefer a more declarative approach, I could certainly implement it similarly to TS (i.e. a populate method and a populated? predicate).

Cheers,
- Clemens

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -5,49 +5,56 @@ module Tire
       include Enumerable
       include Pagination
 
-      attr_reader :time, :total, :options, :results, :facets
+      attr_reader :time, :total, :options, :facets
 
       def initialize(response, options={})
-        @options = options
-        @time    = response['took'].to_i
-        @total   = response['hits']['total'].to_i
-        @results = response['hits']['hits'].map do |h|
-                     if Configuration.wrapper == Hash then h
-                     else
-                       document = {}
+        @response = response
+        @options  = options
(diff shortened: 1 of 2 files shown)


clemens:
I've also fixed the two TODOs for instantiating ActiveRecords with id (avoid the mass assignment warning) and new_record status.

karmi:
Hi, thanks! I'll have a proper look on this.

However, the real issue is to be able to "lazily" perform the search in models? I don't see the benefit of "lazy" here? You aim for not instantiating the documents collection at all? If that is the case, yeah, some (little) RAM/time can be probably saved by this. (Implementation-wise this is perfect, no need to populate/populated hoops.)

Thanks for the `instantiate` stuff!, also, I was searching for a way how to get round this.

clemens:
I have a use case where I only need facets and meta data – not the actual results. Not sure if mine is an edge case but lazy loading usually doesn't hurt.

karmi:
No, instantiating the collection only when needed is fine. Your case is perfectly valid and in _that_ case setting the size to 0 is preferred. I still have to implement the `count` search type, perfect for this.

karmi:
Thanks!, merged in v0.1.14!

PR #45, opened 2011-06-24
Fixed bad calls to MultiJson.encode when logging on 'debug' level

chrisberkhout:
Fixed bad calls to MultiJson.encode when logging on 'debug' level (see https://gist.github.com/6a948b75f27ba140eeca for an example of the error raised by calling incorrectly).

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -275,7 +275,7 @@ module Tire
 
         if Configuration.logger.level.to_s == 'debug'
           # FIXME: Depends on RestClient implementation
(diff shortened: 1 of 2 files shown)


karmi:
Thanks, merged in 0.1.14 -- in fact, I think the proper way to solve it is karmi/tire@2c98b98. That way, with `yajl`, the JSON is still prettyfied, but does not crash where not available.

chrisberkhout:
Ah, yeah. That's good.
It looks like I originally misinterpreted what `:pretty => true` was doing.
Keep up the good work!

PR #47, opened 2011-06-27
don't need to require it twice

jgeiger:
had the same line required twice


karmi:
Thanks!, merged in karmi/tire@2c518c

PR #50, opened 2011-07-01
Adding a size function to index.rb

housepage:


karmi:
Hi,

a) why there are not unit and integration tests attached to the patch?

b) I think what this feature really hints at is something like an `Index#status` method, returning an `Index::Status` object, wrapping `num_docs`, `deleted_docs`, `size`, etc?

karmi:
Hey, @housepage, any news, comments?

PR #53, opened 2011-07-07
Timezone breaking test

jonkarna:
Changed 'Hash should properly serialize Time into JSON' test to account for differing timezone.


karmi:
Thanks!, merged

Issue #38, opened 2011-06-14
ActiveModel integration should search in <INDEX>/<TYPE>


karmi:
Currently, ActiveModel integration searches all docs in the `model_name.plural` index. This prevents us from storing other types of documents in the same index.

The searchable interface should only search for the `"#{index_name}/#{document_type}` documents.

See https://github.com/karmi/tire/blob/master/test/unit/model_search_test.rb#L61-62

PR #54, opened 2011-07-07
ActiveModel integration should search in <INDEX>/<TYPE>

jonkarna:
Added ability to pass type to search. Model passes document_type to search. #38


karmi:
Hi, thanks, merged and cleaned up + amended the logic a bit. Now the interface makes much more sense.

PR #69, opened 2011-08-09
Should force a type to be specified. This bit me in the ass. 

sbeckeriv:
I was very confused why my term search wasn't working. What I created for an index was different then where it was being stored. If there was no default I could have detected this error in my hash earlier. 

Thanks 
Becker

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -280,7 +280,8 @@ module Tire
           document.type
         end
       $VERBOSE = old_verbose
-      type ||= :document
+      raise "Please specify a type" unless type
+      type
     end
 
     def get_id_from_document(document)


karmi:
I feel your pain. Forcing the _type_ to be set by the user, however, isn't very "elastic", is it?

Do you have any suggestion how to document the behavior better?

sbeckeriv:
If the type is not forced then the indexing is arbitrary. This is fine until the coder creates a mapping.  Rages, terms, geo features and other features seem to fail to work only because data is indexed in an unknown type. 

I strongly feel that that raising with out a type is the correct thing to do. If not the default of type would be better if it was you_did_not_set_a_type in stead of document. Still confusing but more descriptive. 

The import method could be rewritten to require an explicit type.

import(:person,[{},{},{}....]) removing the need for the guessing again.

Creating better docs only work if people read them. I skimmed the examples and got started. The code base is very easy to pick up and use. This one part was the only major gotcha I have had. Some people, I am included, read code instead of docs. I did not find the wiki page on how send unsupported searches until after I read the code to figure it out. 

Its better to not allow the user to incorrectly index something then to stay "elastic". 
Thank you,
Becker

karmi:
I understand all your points. I don't agree, however. ES is perfectly fine when you throw random hashes at it. If and when the need for more arises, you have all the means to do so. This, to me, is aligned with the general Ruby approach as well, in my humble opinion...

I'm glad you can get something working quickly. Nevertheless, without reading the docs, you can't and won't do anything moderately interesting with ES. Again, if I can make the wording for the docs or the examples better, I'd love to hear it!

PR #71, opened 2011-08-12
use articles instead of bulk for consistent examples. Also added :type to

sbeckeriv:
use articles instead of bulk for consistent examples. Also added :type to bulk import since there was a mapping declared.  

Instead of the raise. I still think the default behavior is unexpected.


karmi:
Hi, I've cleaned up the message, and rewrote the code so the flow of the examples is preserved. Looks good?

sbeckeriv:
Looks good. Thanks. Everything else is still rockin.

Issue #70, opened 2011-08-12
Tire::Model::Callbacks prevents object deletion


roccoblues:
When I include Tire::Model::Callbacks in a model the records are no longer removed on destroy() from the database. All associations are deleted and also the Elastic Search Index gets update but the final delete statement is never run.

So far I had no luck narrowing it down. Any ideas where to look?

This is with Rails 3.0.9 and tire 0.1.16.

karmi:
Hi,

this is _very_ unexpected behaviour. I've double-checked in an app, and also tightened up the integration tests for this. Could you provide more info about what ORM is this concerning, and possibly some snippets to recreate, please?

roccoblues:
It also fails with a clean model in the same app. Here's my model:

```
class TireTest < ActiveRecord::Base
  include Tire::Model::Search
  include Tire::Model::Callbacks
end
```

testcase in the console:

```
:001 > t = TireTest.new
 => #<TireTest id: nil, name: nil, created_at: nil, updated_at: nil> 
:002 > t.name = 'test'
 => "test" 
:003 > t.save
 => true 
:004 > t.destroy
 => #<TireTest id: 2, name: "test", created_at: "2011-08-16 13:12:34", updated_at: "2011-08-16 13:12:34"> 
:005 > t = Tiretest.find(2)
 => #<TireTest id: 2, name: "test", created_at: "2011-08-16 13:12:34", updated_at: "2011-08-16 13:12:34">
```

the log:

```
SQL (0.1ms)  BEGIN
SQL (1.0ms)  SHOW TABLES
SQL (1.8ms)  describe `tire_tests`
AREL (0.3ms)  INSERT INTO `tire_tests` (`name`, `created_at`, `updated_at`) VALUES ('test', '2011-08-16 13:12:34', '2011-08-16 13:12:34')
SQL (2.2ms)  COMMIT
SQL (0.1ms)  BEGIN
SQL (0.2ms)  COMMIT
TireTest Load (0.3ms)  SELECT SQL_NO_CACHE `tire_tests`.* FROM `tire_tests` WHERE `tire_tests`.`id` = 2 LIMIT 1
```

same test without `include Tire::Model::Callbacks`

```
SQL (0.1ms)  BEGIN
SQL (0.8ms)  SHOW TABLES
SQL (1.9ms)  describe `tire_tests`
[...]

PR #73, opened 2011-08-17
fix for issue #70 (Tire::Model::Callbacks prevents object deletion)

roccoblues:


karmi:
Thanks!! Closed in karmi/tire@418e40a.

PR #76, opened 2011-08-22
Removed ?pretty=true

gjb83:
`?pretty=true` should only be used for debugging. Removed.
(http://www.elasticsearch.org/guide/reference/api/)


karmi:
Yes, and that's _precisely_ why there is `pretty=true` in the debug output of `to_curl` :)

gjb83:
Oops, apologies!

karmi:
No worries, mate :)

PR #78, opened 2011-08-23
[Feature] ActiveRecord integration: index_name can now optionally be passed in to a search. 

bcoe:
It's nice to be able to override the :index_name on a query by query basis. this allows for an ActiveRecord class to interact with multiple indices without race conditions (since the variable index_name is set as a class level variable).

This solves a problem I ran into when an application might have multiple indices being interacted with on the same ActiceRecord model. It's nice to be able to specify an index at search time.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -71,8 +71,13 @@ module Tire
 
           sort      = Array( options[:order] || options[:sort] )
           options   = default_options.update(options)
-
-          s = Tire::Search::Search.new(elasticsearch_index.name, options)
+          
+          if options.fetch(:index_name, nil)
+            s = Tire::Search::Search.new(options.delete(:index_name), options)
+          else
+            s = Tire::Search::Search.new(elasticsearch_index.name, options)
+          end
+          
           s.size( options[:per_page].to_i ) if options[:per_page]
           s.from( options[:page].to_i <= 1 ? 0 : (options[:per_page].to_i * (options[:page].to_i-1)) ) if options[:page] && options[:per_page]
           s.sort do
--- a/test/integration/active_record_searchable_test.rb
+++ b/test/integration/active_record_searchable_test.rb
@@ -15,6 +15,10 @@ module Tire
           t.string   :title
           t.datetime :created_at, :default => 'NOW()'
         end
+        create_table :custom_active_record_articles do |t|
+          t.string   :title
+          t.datetime :created_at, :default => 'NOW()'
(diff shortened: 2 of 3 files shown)


karmi:
Seems like a wrong approach to your problem. Why not just go down one level:

```
Tire.search ['index_1', index_2'] do
  query { string 'foo' }
end
```

And possibly create a class to hold the composed search?

bcoe:
We already have an existing ActiveRecord class, which we have Tire to map the results onto (this class actually uses the MongoMapper mixin as well). I'd prefer to use Tire's mixins for search, mainly because I'm just rebuilding them myself otherwise (and they work great minus the index name).

karmi:
I see. I'd go with `MyClass.search :index => 'other' do ... end` then, and allow `MyClass.search :index => ['other1', 'other2']` then. That makes it more consistent with the DSL approach.

Now, what about the `:type`, then? It should allow to pass `MyClass.search :index => 'other', :type => ['typeA', 'typeB'] do ... end` as well -- otherwise the search won't work as expected.

It still feels a bit weird, and puts another stress on the RDoc documentation (still in the works), but...

Notice, that in the 0.2 release, the casting was changed, all results are `Item` by default, so making a separate component (class, module) wrapping the generic `Tire.search` would probably work very similar. I understand this is much more convenient for the usage you describe. I'll mull it over a bit :)

bcoe:
Yeah, I understand it's potentially a bit ugly. We're bumping into a weird use-case (multiple indices on one model) that your wrapper for ActiveRecord is close to working for, it's not the end of the world if we have to go with our own solution.

karmi:
Two things: first, I smell something weird in using two indices for one model :) That's for sure. On the other hand, the options hash is quite freeform, so yes, it makes sense to expose a flexibility like this. It just needs to be more complete, with the Array stuff, with the `:type` stuff, etc., in my opinion.

bcoe:
We have a multi-tenant application where a tenant might be in one of a number of indices (it's not two indices being punched onto the same model).

Sounds like we're on the same page though :)

karmi:
We are. I still feel you'd be better with some creative use of index aliases, or having an index per user... passing some random index name on search time feels somehow fragile. But I may be wrong and it's _your_ use case after all.

karmi:
Hey Ben, took another approach, the `default_options` lent itself conveniently for supporting this. Also, a full integration test feels redundant for this case...

Could you test it in your app, if it's doing what is should for you?

PR #82, opened 2011-08-28
Support histogram facets

pacoguzman:


pacoguzman:
I suggest the following syntax for the different use cases of histogram facets (maybe applicable to other facets)

``` ruby
# Default histogram using a field 'words'
s = Tire.search('articles-test') do
         query { all }
         facet 'words' do
            histogram :words, :interval => 100
         end
      end

# Custom histogram facet, define the facet with the histogram key
s = Tire.search('articles-test') do
         query { all }
         facet 'words' do
            histogram :words, :histogram => {:key_field => "words", :value_field => "other_field", :interval => 100}
         end
      end
```

[Histogram facets](http://www.elasticsearch.org/guide/reference/api/search/facets/histogram-facet.html)

Other option could be use a custom facet where all the options build the facet like this:

``` ruby
s = Tire.search('articles-test') do
         query { all }
         facet 'words' do
            custom :histogram => {:key_field => "words", :value_field => "other_field", :interval => 100}
         end
      end
```

This pull request implements the first option.

Any opinion about this?

karmi:
Hi Paco, finally merged, thanks! I like the interface you're proposing. I think the duplication of `histogram` is something which we should sort out in the future...

PR #91, opened 2011-09-03
[ACTIVEMODEL] allow to set per-instance custom index name.

abecciu:
This is useful, for instance, to set a custom index name from the model instance attributes before the update_index callback in a safe way.

### DIFF
--- a/lib/tire/model/naming.rb
+++ b/lib/tire/model/naming.rb
@@ -15,8 +15,9 @@ module Tire
       end
 
       module InstanceMethods
-        def index_name
-          instance.class.tire.index_name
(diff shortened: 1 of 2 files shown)


karmi:
Hmm, OK, but how does this work with the rest of the code? `MyModel#index` would still return an `Index` instance based on class level settings?

I understand the reasoning behind that, having to do something very similar previously with CouchDB. I just think this has to be considered across the whole lifecycle of document. Maybe an integration test would find and clarify all the issues?

abecciu:
I'm closing this in favor of pull request #92.

PR #92, opened 2011-09-04
Add dynamic index names

olivere:
Assume the following use cases:
1. You got a historic trail of blog posts that you want to store in ES. Posts from 2011 should get to the `posts-2011` index, those from 2010 to `posts-2010`.
2. You got product catalogs and products (1:n). You want to store products from catalog A in `products-A`, products from catalog B in index `products-B`.

I tried to implement that with something I called "dynamic index names" (in lack of a better word). So instead of specifying the `index_name` as a static string, one could pass a block that will be evaluated at runtime.

```
index_name Proc.new { "posts-#{Time.now.year}" }
index_name Proc.new { |product| "products-#{product.catalog_id}" }
```

I guess something like this might be a basis for index alias management.

Please consider this pull request as an idea/a proposal. There are no tests yet, and there are most probably some issues I didn't think about. Just wanted to see if this is something you'd like to consider.


karmi:
Hi, I like the direction you're going. On the other hand, I think we're mixing several related, but separate _ElasticSearch_ and _Tire_ features.

For the first use case, ES's index **aliases** seem to be the right tool, to me. When you'd periodically updated aliases, outside the app, the behaviour would be consistent across the whole stack. No matter how I put something into and index "named" `posts` (an alias), it will _always_ end up in the index `posts-YYYY-MM`.

Index **templates** would, of course, play a crucial role here, to properly set up all these indices.

Now, I'd of course like support for both in _Tire_. The proposed API for aliases, however, seems to me a bit rough. I'd much rather have something like:

``` ruby
Tire.index('posts').aliases
# => [ 'posts-2011-01', 'posts-2011-02' ]

Tire.index('posts').aliases do
  remove 'posts-2011-01'
  add    'posts-2011-03'
end
```

That's still off the top of my head. The API for templates would be very similar.

Now, to the "dynamic index name" stuff. Thank you for the suggestion and the code; it's precisely what I've been dealing with a lot, in ES, in CouchDB and Couch-related gems.

I think the first case, `index_name lambda { "posts-#{Time.now.year}" }` is very easy to support, I'll add that as soon as possible.

The second use case is more tricky. Unless I miss something, you're evaluating that block in the context of the class. You're doing `Product.search ...`, and there's just no way how to pass an instance with some category set. Of course, it would be possible to do `@category.search`, where it certainly would be possible to search in custom index, like you suggest.

But, that could probably be implemented in a much ore lightweight way by doing something like:

``` ruby
class Category
  # ...

  def search('query')
    self.class.search 'query', :index => self.to_param
  end
end
```

The whole thing could of course be put into a module, which could then be included in the class, etc etc. It would certainly look nice to _my_ eyes, and would be comprehensible, while supporting something like this on the gem level would probably need lots of acrobatics...

What does everybody else think? @floere, @chrisberkhout, @pacoguzman?

olivere:
Thanks for the feedback.

I like your API for aliases (and templates). It's better to use `Tire.index('...').aliases { ... }` or even `Tire.index('...') { aliases { ... } }` than my idea with `Tire.index('...').add_alias '...'`. That seems more consistent to me regarding the rest of the Tire API. 

I didn't think about templates yet, i.e. I set them up outside of Tire. Of course, it would be great to have support for that in Tire, too.

Regarding the dynamic index names:

The first case--when you don't need to pass instances to the index name method--was relatively easy to implement. It will nicely put the documents into the right index.

As you said, the second case is more tricky. There are two issues I considered: Indexing and searching. Indexing looks up the index name at runtime (if it is "dynamic", i.e. a Proc). Search (in the dynamic case) only works if you pass the index name via the `:index` option. Certainly not very elegant... and one of the reasons I said this is just a proposal. Maybe you guys have a better idea.

Unfortunately, it's the second case I need for one of my apps... :-)

floere:
Hey there,

I have a bit of feedback/questions regarding the API:

Why the lambda for `index_name`?

```
index_name lambda { "posts-#{Time.now.year}" }
```

Why not just

```
index_name { "posts-#{Time.now.year}" }
```

?
(In that specific case I also suggest precalculating the year `year = Time.now.year` and then using the bound year inside the block)

I am wondering a bit about this suggestion:

```
Tire.index('posts').aliases do
  remove 'posts-2011-01'
  add    'posts-2011-03'
end
```

Why the departure from "simple" Ruby?

If

```
Tire.index('posts').aliases
```

is just an array, it would imho be better to treat it as such:

```
aliases = Tire.index('posts').aliases
aliases.delete 'posts-2011-01'
aliases << 'posts-2011-03'
```

(or use whatever Array method you like best…)

I do like @karmi 's suggestion for searches and find it elegant. Or, especially, readable – it is very clear what is done there. I presume this is a special case and I personally would refrain from moving it into the library. (But I might be wrong, of course)

I am intrigued by the hand-standish way this needs (?) to be solved. In certain other search engines _cough cough_ this is not worth talking about ;) (OTOH no incremental index updating, so I'll shut up now)

But seriously: What is the actual need for multiple indexes here?

olivere:
Thank you for your perspective, @floere.

Last things first :-)  I apologize if my tone sounds rude or like I want you guys to implement exactly what I need for my projects. That's certainly not my intent. I made a proposal of something I found to be of general interest. You might disagree, propose a different solution, or don't care. I'm fine with that, it's just the way it is. :-)

Having said that, here's my train of thoughts for why this might be of general interest:

Suppose you have a data model of catalogs, products, and groups. Catalogs have products, and groups have access to certain catalogs, i.e. not all groups have access to the same set of catalogs. You want groups (of users) to search for products. So you build an index on the products and filter by the catalogs by the group. There we are...

Now catalogs come and go, so you need to remove some products from the index. You could use delete_by_query, but that probably has a negative impact on performance (I didn't test that personally, but found [at least one comment](https://groups.google.com/group/elasticsearch/browse_thread/thread/781d2148417d99eb/37d481973658d937?lnk=gst&q=delete_by_query#37d481973658d937) from Shay to refrain from using delete_by_query on large indexes). 

So, how can we combine fast search with simple and fast management of catalogs? My solution was to create one index per catalog and use aliases to take care of the (group) views. Creating a new catalog means creating an index and adding it to the right aliases. Removing a catalog simply means dropping the index. Searching as a group basically means finding the right index name--I don't even have to bother about filtering.

If you think this might be a valid approach/use case, you have to figure out a way to put products into the right index. Hence the approach to evaluate the index name at runtime, e.g. use a unique catalog identifier to build the index name.

I'm not sure about it but I guess there are similar data models where this approach could prove to be beneficial. You decide :-)

**Index name**

Using a block for the index name instead of a lambda looks better to me.

**Aliases**

Regarding the aliases syntax, I'm not sure. Code like this

```
aliases = Tire.index('posts').aliases
aliases.delete 'posts-2011-01'
aliases << 'posts-2011-03'
```

is more Ruby-ish, while

```
Tire.index('posts').aliases do
  remove 'posts-2011-01'
  add    'posts-2011-03'
end
```

is more like the rest of the Tire syntax, e.g.

```
Tire.index 'articles' do
  delete if exists?
  create
  store ...
  refresh
end
```

**Search method**

You guys are right. Passing the index name to the search method already exists. I suppose people wrap the search method with some custom logic anyway, so why bother.

redbeard:
@olivere, a small comment regarding the solution you suggested for the performance hit of the delete_by_query.

This is not disimilar to the SQL world in which a delete is a costly operation (blocks need to be moved / de-fragmented). The solution is usually to have an `active` flag on an entity which you can then easily filter by. Set it to `false` for a 'deletion'. 

This should not be not as costly on delete as the space for the flag is already on the disk in the index. 

Seems to be a simpler solution rather than having to maintain multiple indexes, or am I misunderstanding your problem?

olivere:
@redbeard, I appreciate your suggestions. Thank you.

I assume marking documents with `active: false` shouldn't be as expensive as deleting them right away, am I right? To get rid of the inactive documents I could rebuild the whole thing, or defer `delete_by_query`and `optimize` to times of low work load. Hmm... dropping and creating indices is very fast from my experience, so I thought I could get the "free lunch" ;-)

I guess I have to do much more testing to get a better idea of the whole picture. The "index juggling" solution is working quite okay right now, but I didn't get into the millions of documents. We'll see.

karmi:
Hi, thanks for all the feedback.

First thing, I think that what @olivere is after is a very realistic use case, based on my experience. Frequently, you want something like storing documents in an index per user (think bookmarking service), or some kind of lightweight sliding window done inside the app, etc. Or your use case with catalogs. So yes, supporting workflows like these is crucial for any library worth its salt.

Regarding the dynamic index name, yeah, that's a given. As for:

> Why the lambda for index_name?

Well, it depends on how “dynamic” you want the index name to be. A block is evaluated once, when the class loads. A lambda would be evaluated repeatedly, on every `index_name` call. You could of course memoize the result of the calculation, as @floere suggests. Of course, a decent implementation could easily support both of these :)

Regarding the aliases syntax, yeah, that's one of the things we were discussing with Florian, a lot. I think I'd go for the more DSL-ish syntax, so it's symmetric to the rest of interface. Of course, you can always use the gem on the low-level, ie. do something like `Tire::Index.new('something').aliases << 'something-else'`, etc.

Regarding the `delete_by_query`, yeah, dropping (or closing!) an index would be almost instantenous. If you can do it like that, go for it! (By the way I'd be more concerned with unpredictability then performance.)

olivere:
Fantastic. Just let me know if and how I can help.

floere:
@olivere, you don't sound rude at all, quite the opposite. I have to say I am not extremely invested in the ES/Lucene world anymore, but enjoy API discussions and this one so far. @karmi has probably added me for the API side of things.

### Indexes

I should clarify:
Earlier I was wondering why you try making index names dynamic, when Lucene already offers dynamic functionality using a combination of field values and filters. (Something along the lines that @redbeard suggested)

I.e. have a year field with values 2010, 2011 etc., then only returning results with year 2011.

Personally, I'd index the same data in the same index, to enable Lucene to make optimizations etc.
(I've talked about it regarding user-specific results here: http://florianhanke.com/blog/2011/08/31/picky-case-study-restricting-results.html)

However, I can see why using separate indexes can be advantageous. (Probably a bit easier to think about and also more modular data, on disk, for backup etc. and other reasons)

How is performance when having many (> 1000) indexes for as many users?

### Lambdas

> A block is evaluated once, when the class loads. A lambda would be evaluated repeatedly, on every index_name call.

@karmi, I probably misunderstand this.

If I don't:
How or when a block/lambda is evaluated is up to the framework/method etc. the lambda/block is passed to.
If the lambda is evaluated/called, there are some differences (e.g. you can use `return`).

A quick illustration:

```
def look_a_lambda lam, &block
  p lam.call
  p block.call
  sleep 1
  p lam.call
  p block.call  
end
look_a_lambda(lambda{ Time.now }) { Time.now }
```

Or are you telling me how Tire would handle the block/lambda?

I probably wouldn't support lambdas at first. The use case for them is too thin, and blocks are more elegant, imho.

> You could of course memoize the result of the calculation, as @floere suggests.

I actually wasn't suggesting memoization (in a strict sense, but maybe I am being too … picky ;) ), but simply by using the local variable binding :)

### Blocky design

I can see the preference on using blocks. If the low level option is offered, why not :)
(Personally, I think it's a fine line – and it's hard not ending up in LEGO land)

Btw, I am intrigued by `Tire::Index.new('something').aliases << 'something-else'`. You say it's low level Ruby, while to me it looks like a fluid interface, where the `<<` method returns the index itself. Which is it?

abecciu:
+1 for dynamic index names. I desperately need that feature. Index aliases would be a nice addition too.

olivere:
@floere, thanks for clearing this up. I mislike negative vibes. Happy now :-)

**Index API**

Let me add some remarks regarding the API.

I'd like to remind all of us that the Aliases API is more than just adding or removing a single index to an alias ([Aliases API](http://www.elasticsearch.org/guide/reference/api/admin-indices-aliases.html)). You can add and remove several aliases in one batch, use aliases as filtered views, and add routing. We should take this into account when fixing the final API.

Let me also express my views of the general API discussion:

While I generally like mimicking a driver/wrapper gem the Ruby-ish way, it might lead to problems. Let me explain:

```
aliases = Tire.index('posts').aliases
aliases.delete 'posts-2011-01'
aliases << 'posts-2011-03'
```

This seems very straightforward at first sight. However, `delete` and `<<` do communicate with ES behind the scenes. While this might be debatable in the example above, I tend to be explicit when something's going on behind the scenes (in this case: communicating with ES). If you don't make this very clear, people call your code without thought and think that something very lightweight is going on (like adding an element to an array). I agree, this is a very fine line. But @karmi's approach of using a `Tire.index(...) { ...}` block really does help to make this very clear: You are doing something in the context of communicating with an Elasticsearch system. So beware! ;-)

Furthermore, how do you expect Ruby style with the additional Aliases API features. While adding in a batch might be

```
aliases << ['posts-2011-01', 'posts-2011-02', ...]
```

I do not know of a good Ruby idiom to both add and delete in one batch. Same for filtering and routing.

Long story short: Favour the block syntax (while you might still enable users to use the more Ruby-ish stuff).

**Performance**

To be honest, I don't know how performance will be with e.g. 1000 indexes added to one alias. If there are issues, they are related to Elasticsearch, not Tire. But you're right, I should check this, although I expect the number of views to be in the tenths.

When it comes to drivers/wrappers, I like them to be transparent to the underlying technology (Tire -> ES). If it's easy to prevent havoc at the driver level, do so. But in the end, people will always find a way to shoot themselves into the foot. There's just no way around that. ;-)

floere:
@olivere, good to hear :)

### Index API

I didn't know they communicated behind the scenes. So the block now makes a lot of sense to me, thanks for clearing that up.

### Performance

If there are general problems with using many indexes, I'd at least "syntax pepper" it, as a reminder that you ought to know what you are doing. Same argument as the one for using a block, in the topic above, really.
(I have to say, either you want it to be transparent, or not ;) Personally, I am for semi-transparence. As transparent as possible, with added syntax sugar/pepper to guide usage behavior. So far, people like it :) )

Anyway, that said, I'm neither one of the maintainers, or have anything to say, but just a passersby, so take all with a big fat grain of salt.

### Aliases

I have to admit, I probably misunderstood the whole concept of aliases. Can someone clear this up for me?
From the name "alias", I assumed the alias would be in a 1-1 relationship to the original alias.
But now, from your writing, it appears that it is a n-1 relationship, and a search on the index will search on all of the indexes.
(I admit I could have seen this earlier, but it only clicked now)
If this is true, is "alias" a good name?

Cheers

floere:
Re "Aliases": Feel free to tell me to RTFM, by the way :)

olivere:
From the ES guides:

> APIs in elasticsearch accept an index name when working against a specific index, and several indices when applicable. 
> The index aliases API allow to alias an index with a name, with all APIs automatically converting the alias name to the
> actual index name. An alias can also be mapped to more than one index, and when specifying it, the alias will
> automatically expand to the aliases indices. An alias can also be associated with a filter that will automatically be applied
> when searching, and a routing values.

or (for @floere): [RTFM](http://www.elasticsearch.org/guide/reference/api/admin-indices-aliases.html) ;-)

floere:
Thanks, @olivere! :)

Personally, I find "alias" very misleading. I'd have used "grouped indexes" or similar, but there you are.
As an inspiration, in Picky you group indexes in searches, like so `Search.new(index1, index2, index3, ...)` which is separate to the "aliasing".
Naming things is hard, but I feel the designer of ES gave up a little quickly ;) I, for one, was misled. But maybe I am the only one.

To be clear, the first part is fine (http://en.wikipedia.org/wiki/Aliasing_(computing)), but the grouping, not so much. All imho.

Thanks for the discussion, it was very enlightening.

karmi:
Thanks for the discussion, guys. Maybe too long for Github Issues, but great :)

Re: **aliases**. I think the name `alias` is apt and fitting here, really. This is about the excellent design ES's API has. I don't think @kimchy "gave up too early" here, at all.

You can do `Search.new(index1, index2, index3, ...)` with ES, of course. See the http://www.elasticsearch.org/blog/2010/02/12/yourdatayoursearch.html blog post for a great "Amazon-like" example.

There's indeed a great deal of architectural and performance benefits you gain from separating your data into separate indexes. One use-case is the "sliding window" one, which I always bring up -- please meditate on that a bit more :) Another one is the "let's put products into separate indices based on category". Yet another one is the "every user has her own index" (with an obvious limit to scaling).

Now, aliases do play with almost all of these use cases. The discussion was not entirely clear so far and I am to blame, mostly.

When we consider two indices, "index-A" and "index-B", these are two `index` entities.

We can create _another entity_, an `alias`, for these two inidices. It is a _single_, separate entity, called `alias`. It can be an „alias”, or pointer, or reference to one or more “real” indices. It _aliases_ them. It's just a nickname, a _nom de guerre_, an A.K.A. for one or more “indices”. You can use their “real names”, or you can use this _alias_. It depends on what is more comfortable for you, in a given context.

What we gain from that is architectural leverage: eg. when more applications in one "project" or "stack" or whatever perform searches, they can do that against the `alias` entity, not the `index` entities. We can close or delete individual indices without affecting the stack. All of this can be managed completely outside of the application (which could be a _very good thing_, even though many Rails developers don't think about it like that).

So, yes, _alias_ is a good name. It's a different concept then when you do `ln -s`, of course, where you have 1:1 correspondence. But when you say “[G7](http://en.wikipedia.org/wiki/G7) decided that ...”, you're using an _alias_ for a ever changing list of countries. A nickname, a _nom de guerre_, an A.K.A. An _alias_.

It is just not clear from the code examples I dumped here originally (`Tire.index('posts').aliases { add "whatever" }`).

In this example, the relation is reverted -- we are in fact working with a "posts" `alias`, not a "posts" `index`. However, with certain exceptions, an `alias` behaves similarly as `index`. So my original intention here could have been _either_ a big confusion _or_ an inspired thinking on how to make the API fluent :) We'll see if/when somebody gets to implementing it for real.

> Btw, I am intrigued by Tire::Index.new('something').aliases << 'something-else'.
> You say it's low level Ruby, while to me it looks like a fluid interface,
> where the << method returns the index itself. Which is it?

I know. I think the thing is `<<`. This just smells “too much Ruby, too little DSL”, to me. Of course, you can write:

```
class Array; alias :add :push; end

aliases.delete 'posts-2011-01'
aliases.add    'posts-2011-03'
```

But in general, I prefer the blockish DSL ways -- so far it's pleasant to use inside applications or one-off scripts, etc. It's very readable even for people with little Ruby knowledge... Unless you descend into fifth level of the blocks, such “declarative” notation feels great.

Of course, for lots of use-case, a more “imperative” notation would be more appropriate. _Tire_ does not hold that against you, at all:

``` ruby
    require 'rubygems'
    require 'tire'

    include Tire

    index = Index.new('articles')
    index.delete
    index.create
    index.store :title => 'Test Article', :tags => ['search', 'ruby']
    index.refresh

    search = Search::Search.new('articles')
    search.query { string('title:T*') }
    search.filter :terms, :tags => ['ruby']
    search.size 1
    # ...

    p search.perform.results
```

I'm not sure why anybody would choose to “talk like that”, generally, but I am 100%sure  there _are_ occasions you'd like to talk _exactly_ like that :)

chrisberkhout:
Hi all,

Very interesting so far!

Sorry this is so long! Maybe this discussion needs to be split into several different issues.

### The term 'alias'

@floere I think I can see where you're coming from with this being confusing. However, I had no worries with the jump from aliases to 'groups' because I was thinking:
1. Wherever you can use an index, you can also use several indexes.
2. Wherever you can use an index, you can use an alias that is a pointer to an actual index (or several of them).

I probably just happened across the right part of the ES docs at the right time.

In any case, Tire should obviously use terminology that's consistent with ES.

### Alias API

@karmi, I also spotted the inversion of `alias` and `index` in the example. I suggest something like:

```
Tire.alias('posts').targets do
  remove 'posts-2011-01'
  add    'posts-2011-03'
end
```

I do agree that managing it outside the application is probably a good idea. I'd probably still want to use Tire to do that.

### Blockish DSL vs 'imperative' Ruby-ish

I can see both sides of this. 

The DSL way is good when you're writing something static. I think you want to talk the 'imperative' style when building up queries dynamically.

I actually did my dynamic query building with the block syntax (mainly because that's what I saw in the README), and it works...

```
search = Tire.search(ES.index_name) do
  size 1000
  query do
    if boolean_parts.count > 0
      boolean_parts.each { |part| part.call(self) }
    end
  end
  FacetParts.for_query(this_query).each { |part| part.call(self) }
  sort_part.call(self) if sort_part
end
```

...but it requires jumping through hoops for no good reason. I'll probably try to refactor it to imperitave style. A mention of imperative style in the README could be good. I didn't realise I could do it that way until I looked more at the Tire code.

I may be opening a can of worms here, but I've also been thinking, something like [Arel](https://github.com/rails/arel) might work even better. Say, like this:

```
s = Search::Search.new('articles').
      query_string('title:T*').
      filter(:terms, :tags => ['ruby']).
      size(1)
```

My reservation is that perhaps that's getting too much into applications (ie. dynamic query builder), rather than sticking to being an API wrapper. Thoughts?

### Dynamic index names for storing and searching

The main story about aliases and multiple indexes is that you use a fixed alias, then separately manage it to resolve to the right index(es). For that case, being able to dynamically choose an index is not important.

You would want the dynamic functionality if you're trying to choose the index based on attributes of an object you're storing, or if you're querying a subset of indexes, chosen based on things not modelled in your alias->index setup.

As said, you can do that already give `Tire.search` a dynamically generated index name, but being able to use a block in place of a static name in models `include`ing `Tire::Model::Search` would be a simple and good enhancement.

Regarding the trickiness that @karmi mentioned in his initial post, what about of two index name blocks: one for querying (which takes no parameter), and one for the index for storing (that takes the object to be stored as parameter)?

### Index templates

Yes! These are going to be important if you're creating lots of indexes on the fly.

### Blocks vs lambdas

For a second I thought that this was something to do with local varaibles defiend before the block/lambda, but then I did a [refresher on blocks and lambdas (and procs)](http://www.robertsosinski.com/2008/12/21/understanding-ruby-blocks-procs-and-lambdas/), and they seem equivalent to me. Except that with blocks you only get to attach one of them to a method call, and with lambdas you need to get the number of parameters right. If the method takes a block I can choose to create and save it a lambda and pass that to the method, if I want.

If there's more to this I'd like to know.

floere:
### Sliding Window

I understand your sliding window examples. It is about separating the index data space somewhere else that just by using a (virtual) field.

Instead of putting everything into one index, and declaring for each indexed document a (possibly virtual) field like "user", with values like "1", "2", etc. the data is split over several indexes. So instead of making the cut at the field level, you make it at the index level.
This is the gist of it to me.

Now this is going to have implications for different searches.
I assume as well that splitting things over several indexes, and then not having to define the added filter, e.g. "user:2", for a user-specific search (where you only search for this user's data), will speed things up

An aside: However, that just might not be the case. Perhaps, ES indexes "user 2" beautifully, and this index lookup is actually – almost – the same speed as selecting the index.

Another implication is that searches that go over the whole index space (i.e. over all the user's indexes) will most certainly be slower.

What I am trying to say here:
By deciding to use n indexes instead of just one, you make a conscious decision that you know what you are doing, while taking a bit of optimization options away from ES.

I am unsure what my opinion on it is, though. First of all, I think it is a good thing that Tire offers the possibility to do so.
For example, if you are certain that you won't do many searches that use all of the user indexes, but mostly, or only on single user data, then by all means use separate indexes.

What I am actually trying to say:
While it might be extremely clear and obvious to you that using separate indexes is the "one true way" (tm), I just wanted to remind you that everything comes with a tradeoff involved. And that choosing which way to go might not always be so painfully obvious, but needs knowledge about how ES works.

So, all in all, yes: I think it's good that Tire offers this feature.
Just wanted to say that really thinking about what one is doing is a good thing, as always. (Yes, in this specific case the performance etc. difference might be tiny, it might be bigger in other cases)

### Outlandish API idea

During the above, it occurred to me how nice it would be if ES could do the separation itself.
Do you guys _really_ care if it's in separate indexes or not?
I don't at all. Assuming, of course, that I never fiddle with single indexes e.g. on the disk.

Right now, I am thinking that it would be nice if ES did the actual optimization. Perhaps supported by the user saying "Hey ES, you know, I'll never be searching over all users. I'll only be searching in each user's index space separately."

### Alias – clarification

If you're pulling @kimchy into this discussion, I suppose I should clarify my comment.

### Preface

I went ahead and read http://www.elasticsearch.org/guide/reference/api/admin-indices-aliases.html. In that context, the aliases make sense. In Tire it still doesn't, to me – if you're interested in my reasoning, see below.

### Ok, go

I, too, think that "alias" is very fitting. And yes, it is great that an index can have multiple aliases.
Let's take me, for example. I am called "Florian". You can use "Flöre" too, to call me. I will listen to this alias.

Ok, so we have now established that an alias is a piece of text that maps to a thing. Like a reference. So, name -> thing.

But friends also call me "Flo" or "Flori". I will listen to each of these nicknames, or pseudonyms, aliases.

Let me put that in code:

```
florian.aliases = ['Flöre', 'Flo', 'Flori']
```

This is how I would define the aliases for myself.

So we have multiple pieces of text that refer to a thing, "Flöre" -> florian, "Flo" -> florian, "Flori" -> florian.
I call these a group of aliases, or "aliases".

Basically, the semantics of an alias is a text pointing to a thing.

Now, in the case of Tire – assuming I have not horribly misunderstood – we get a whole different semantic:

index.aliases = ['index_alias1', 'index_alias2', 'index_alias3']

This now represents the _other_ direction. This is actually the index pointing to a list of index names.

Remembering the florian's aliases above, can you understand how I'm quite confused by the API.

I guess I missed your last paragraph, above. 

> In this example, the relation is reverted -- we are in fact working with a "posts" alias, not a "posts" index.
> However, with > certain exceptions, an alias behaves similarly as index. So my original intention here
> could have been either a big confusion or an inspired thinking on how to make the API fluent :)
> We'll see if/when somebody gets to implementing it for real.

So it appears you see the (or a related) problem as well?
I have a suggestion:

### Aliases vs. Grouping/Bundling

Again, I like the idea of having aliases,

"name" -> thing

but I think it's being overloaded with semantics, as soon as

index -> list of index names

I love the idea of having an index that bundles multiple indexes. But I still think the naming is off, since the used word, "alias" is being used with two very different semantics.

An aside regarding G7: I just recently discussed this with a linguist at my lab. Yes, it is an alias – for "Group of 7". But it is actually referring to the single instance of group, not the list of members. And yes, the members of the group are exchangeable. And yes, "G7" refers to the group. But your argument mixes two things, which is a nice summary of my confusion:

"G7" -> Group of 7 -> list of each member
alias -> bundle frontend -> list of bundled things.

So my suggestion is not to mix these ideas and instead make them separate entities. Something like:

['Alias1', 'Alias2', 'Alias3'] all point to IndexA which bundles Index1, Index2, Index3.

So, in Tire, this would be:

```
Tire.index('posts').bundles('posts-2010', 'posts-2011').aliases('theposts', 'p')
```

### It all comes down to

Using separate names for semantically different concepts.

### In closing

Phew :) Thanks for reading all that and yeah, @karmi, probably it's an idea to just go ahead for now. I still hope my point came across, and that it _might_ be potentially confusing.

Please also note that the amount of written text above doesn't relate to the importance – but more about how much text is needed to be clear :)

floere:
"If there's more to this I'd like to know."

@chrisberkhout: Yes, the return issue, as I say. Consider:

```
def call_the lam
  lam.call
end

p "lambda"
call_the lambda { return }
p "Proc.new"
call_the Proc.new { return }
```

Cheers!

chrisberkhout:
@floere Yep, thanks. Got that one from my reading. Just forgot to mention it
again because I think it's not really relevant to the DSL issue.

floere:
@chrisberkhout It only is if the API uses the return value from the block. (Which it probably won't afaics)

chrisberkhout:
@floere Oh yeah, that's a good point. Still have options there though:

```
# block that uses "return" value
def block_eater
  puts "block_eater got: #{yield}"
end

# ways you can't return stuff
block_eater { return "return hello from block" } # => LocalJumpError: unexpected return
block_eater(&(lambda { return "return hello from lambda" })) # => LocalJumpError: unexpected return

# ways you can "return" stuff
block_eater { "hello from block" } # => "block_eater got: hello from block"
block_eater(&(lambda { "hello from lambda" })) # => "block_eater got: hello from lambda"
```

Hmm... getting off topic here.

floere:
@chrisberkhout: Yes, by using `&` on a lambda object it's downgraded to a simple block and an explicit `return` can't be used anymore. So if the API method wants to accommodate `return`s in blocks it needs to accept lambda objects in its params, i.e. `def method(lam)`.

You're right, we're getting off topic. We probably were doing so starting at comment 1 ;) (But hey, inspiring discussion!)

olivere:
It all started with a feature request... now: semantics and Ruby subtleties... OMG what have I done... ;-)   (no really, great input from anyone)

karmi:
Hi all,

> Maybe this discussion needs to be split into several different issues.

Yeah, I think we have filled like three separate discussion threads already. Isn't this some kind of elaborate plot to lure me into creating a Google Group, guys? :)

> `Tire.alias('posts').targets do ...`

I am aware of that. I _think_ however, that starting with the `Tire.index` could eventually yield something. As said: I am either totally confused or inspired. In ancient times, they have tend to look at these as being similar :)

> A mention of imperative style in the README could be good.

Yes! Added that, you're right that it could be very useful.

> I may be opening a can of worms here (...) something like Arel might work even better.

Who knows :) I am kinda hesitant, here.

First, there's a great project by @clintongormley, [SearchBuilder](https://github.com/clintongormley/ElasticSearch-SearchBuilder), doing just that. I, obviously, see the benefits of an approach like this, and it would probably be “easy” to support Arel-like syntax for searching.

The reason I'm kinda hesitant is that I have always tried more to expose the ES API in a very Ruby, and hopefully elegant, pleasant manner, then to invent radically new interface. I have certainly tried to make ElasticSearch more “approachable” for a regular Ruby/Rails developer, but a syntax like this would probably go a bit too far. On the other hand, I am quite sure there are benefits for something like that.

> Blocks vs lambdas

Again, the confusion is entirely my fault. You can, of course, just _store_ the block defining the index name, and _call_ it later, in the `index_name` method.

> By deciding to use n indexes instead of just one, you make a conscious decision that you know what you are doing (...)

Yes!, precisely. I wouldn't put stuff into “one index per user” if I expect there will be thousands or more users. For the “sliding window” scenario, it would work rather well, because then I can optimally deal with potentially expensive operation such as “delete all records older then 3 months”. It's really very similar to eg. [PostgreSQL partitioning](http://www.postgresql.org/docs/8.3/interactive/ddl-partitioning.html).

> `index.aliases = ['index_alias1', 'index_alias2', 'index_alias3']` (...) This now represents the other direction. This is actually the index pointing to a list of index names.

Yes, that is where I have started the confusion, by inverting the thing around. I'd just say once again, that something in my mind tells me it _could_ work. Ie. the `create` method would then create an alias, not the index, etc. Something in my mind tells me as well, obviously, that it could end up being an unbelievable mess :)

Once I recover from all the wild ideas, I'd have a shot at the dynamic index names the OP was after :D

karmi:
So, I finally got to implementing the “easy” variant of dynamic indices, the class level one:

``` ruby
Article.index_name { "articles-#{Time.now.year}" }
```

The instance level one, summed very well in the example code from @olivere, remains “tricky”.

I have, however, find a promising approach to this. The right interface is really obvious:

``` ruby
@blog.posts.search 'something'
```

This way, at least in ActiveRecord, we could have use the scope information. I had, however, _not_ found a way implement that for real.

The funny thing is, such syntax is absolutely valid, and it works -- since ActiveRecord passes the call to `search` via method missing to `Post`. The scoping is lost, nevertheless.

If we may get around that, the interface for AR models would be very fluent. I worry, still, that it would not be portable to ActiveModel implementations for other databases, though.

One last thought: maybe we are simply stretching the gem too wide here. Why not just overload the `MyModel.index_name` and `MyModel#index_name` methods, and be done with it. Yes, ugly, but kinda “in plain sight”. No magic and convoluted infrastructure.

All that said, I am the first one, ironically, who would use that in production, immediately. We are “scoping” searches like that to user-bound indices in one of my apps from the humble beginnings with CouchDB and CouchDB-Lucene...

olivere:
Sorry for the long delay... I've been moving and spend some days off.

First of all, thanks for the commit. I'm with you on `@blog.posts.search 'something'` and I also like the block-passing to `index_name`. However, I think we're still not there with instance-based persistence and alias management, am I wrong? Or am I just too dumb to figure out how to get Tire to decide which index is used for storage on create/update time.

So short question: Are you still planning to add features for handling aliases and store entities on instance-scope? Or, if it's already possible, provide a short example of how to achieve that?

karmi:
@olivere: You can do instance-based indexing/persistence just fine, the only thing you have to do is overload the `Article#index_name` (instance!) method. Try it:

``` ruby
class Article
  def index_name
    self.author.fullname.parameterize
  end
end
```

You cannot simply do instance-based _searching_, since `Article.search` is class method and all that. `@blog.posts.search 'something'` is very, very nice, I'm just not sure how to properly support/implement that, for ActiveRecord _and_ ActiveModel.

As for the aliases, I can definitely rule out focusing on that -- I'd much rather focus on adding the `:timeout` option (see separate issue), the `search_type=count` option for speeding up facets, figuring out how to support multiple filters, etc etc...

karmi:
@olivere: I'm sorry, it's more tricky than that and than I'd like. This does _not_ work as expected:

``` ruby
require 'rubygems'
require 'tire'
require 'active_record'
require 'sqlite3'

Tire.configure { logger STDERR }

ActiveRecord::Base.establish_connection( :adapter => 'sqlite3', :database => ":memory:" )

ActiveRecord::Migration.verbose = false
ActiveRecord::Schema.define(:version => 1) do
  create_table :articles do |t|
    t.string   :title
    t.string   :author
    t.datetime :created_at, :default => 'NOW()'
  end
end

  include Tire::Model::Search

  def index_name
    author.parameterize
  end
end

one = Article.create! :title => 'Test custom index', :author => 'John Smith'
two = Article.create! :title => 'Test custom index', :author => 'Mary Smith'
```

Tire needs a patch in [_callbacks.rb_](https://github.com/karmi/tire/blob/master/lib/tire/model/callbacks.rb#L21-22). I'll look into that laters.

olivere:
Thanks @karmi.

I can handle the aliasing stuff, perhaps even submit a pull request for you to review (based on what we discussed in this issue). The only showstopper I've got is the instance-based persistence.

karmi:
@olivere: I'm not sure at the moment how to fix the logic. The easiest would be to just call `update_index` when it's defined in the model instance, but that would also cross the `tire` “proxy” boundaries... I'll have to mull it over a bit. The infrastructure itself definitely supports that. It's just about deciding where to “make the cut”.

olivere:
@karmi: Sure, I think I know that feeling when you're sure something is definitely doable but it doesn't seem right. Better take your time and do it "the right way" (tm). It's not really time-critical right now, as I'll can focus on other parts of the application. Just let me know if I can help somehow, write tests, whatever.

olivere:
@karmi: Any news on this?

karmi:
No. It's highly probable I'll bump into this same issues again, with a different application. So I'l have a real problem and real pain to extract the solution from it.

olivere:
Wonderf... oops, sorry ;-)

karmi:
Haha :) If you need to pingpong ideas for workaround, catch me on IRC. But I'm sure we will have some support for these dynamic things soon.

olivere:
I still rely on my aliases branch which works okay but it's a bit off-track of the original tire repository. I prefer to be on the maintainers master branch because everything else is just a PITA after a while. 

Keep on the good work.

olivere:
Just a thought experiment: We agree that searching can be done with something like `@blog.posts.search '...'`, so we're only left with designing the persistence stuff. Then again, if all we need to do is to choose index/index name on save, isn't that (a very simple form of) sharding?

If it is, it may be an idea to have a generator/facade which simply returns the static index name in the general case but may be overridden to return something more complicated (e.g. based on the instance). That idea could take us even further without complicating the code too much or drifting away into edge cases. What do you think?

karmi:
Guys, not to wake sleeping dragons, but there's some relevant code in https://github.com/karmi/tire/commit/aliases as of yesterday :)

olivere:
So I can now remove that `tire_aliases.rb` from my projects that'll monkey-patch Tire to include `add_alias`, `remove_alias`, and `aliaes` methods? Lovely. Thanks.

rathgar:
_blows the dust off..._

Thanks for the aliases stuff @karmi, that's really handy.

We do however, have a use case that requires us to dynamically name the indices. We have many customers with 10k to millions of documents which we currently have all in one index for each type. We'd like to be able to rebuild a customer's index from time-to-time for various reasons and of course when a customer leaves it's easier and safer to just drop their indices. Did you get far @olivere?

We can create aliases with filters to just their documents but that doesn't allow us to rebuild/drop individual customers' indices.

When any document action takes place the `index_name` would need to be calculated within the context of the document. I've tried overriding the instance method `index_name` but it seems that it is currently called on the class.

I'm happy to help move this 'feature' forward if anyone is able to give some pointers as to where best to start. It would, obviously need to function without cost to the current method. 

If no one else would be interested in using this, I'll monkey patch and gist/fork but will probably still need some guidance on the best way forward.

olivere:
@rathgar, I will try to explain what we did to segment our big index into smaller ones. It's been quite a while since we did this, but IIRC it was quite easy and robust once we got it working.

We used to have a big `products` index which is now indexed by tenant, like `products-1`, `products-2` and so on. So our "dynamic index name" is made up by `products` and the ID of the tenant. We like this not only to automatically ensure tenants only see their data, but also because it's very easy to drop a tenant: Remove some rows from your data store and drop the ES index. No need for `delete_by_query`...

On creation time, we made sure that an ES template is applied to all products indices, so we automatically get the same schema. When importing data we "dynamically" determine the right index name based on the `products` prefix and the tenant ID (saved in the User model). Once you know the index name, you can limit all further ES actions (search, bulk import calls etc.) on that index name. 

When searching, we know the users' tenant, so we can also build the right index name and pass it around to the various Tire methods. IIRC we had a `current_user.search_products(...)` convenience method that did all the right things. 

You can drive this even further, e.g. when some users get super powers and are allowed to view two or more tenants. This is where Aliases come into play... when users get access to a new tenant via some `UserTenant` model, just set up an alias as you create the `UserTenant` and drop the alias if you destroy the `UserTenant` model.

I hope you get the idea... feel free to ask again if it's too confusing. I apologize in advance...

rathgar:
Thanks for your feedback @olivere, that's very useful and seems to be almost exactly the same scenario as ours. I'd be grateful if you would be able to clarify a couple of points. _I should have mentioned before that we're using `ActiveRecord` (Rails v3.2.6)_

We're also wrapping search so I'm happy that we can do that against the appropriate `aliases`. However, if we drop our current indices and import each customer's documents into its own indices (one for each type):
- Is it correct to assume that during `ActiveRecord` updates and deletions, Tire will still look for a `products` index and that we would need to create that as an `alias` pointing to all of the `products-1` ... `products-n` indices?
- Similarly, when new records are created, how will `Tire` know which client index to add them to?

These are the reasons I assumed I needed to coax `index_name` to be 'dynamic'. We couldn't do future index additions via bulk import as the models get updated and created very frequently and new/changed objects need to appear correctly in search immediately.

Sorry if any of this is already covered elsewhere.

olivere:
(I hope it's okay for @karmi to use this issue as a kind of support thread... :-))

To your questions:

**Searching:** I'm pretty sure that you can use either alias or index name (correct me someone if I'm wrong). IIRC the alias name will just be expanded to the aliased index names. I'm not sure about the limits of this approach (it might not be a good idea to e.g. have millions of tenants); in our case, the number of tenants is quite manageable.

**Indexing:** We perform background indexing via the Java-API so we are not dependent on Tire/ActiveRecord callbacks. However, I'm pretty sure you can just disable the callbacks and manage the indexing process yourself. After all, Tire does just that in [Tire::Callbacks](https://github.com/karmi/tire/blob/master/lib/tire/model/callbacks.rb). @karmi wrote about it in the README (lookup the section on CouchDB/RabbitMQ), so I suppose it's a common use case.

**Reindexing:** Did you take a look into Tire's [reindex method](https://github.com/karmi/tire/blob/master/test/integration/reindex_test.rb#L28). I'm not sure if you can rename an existing index. But you can achieve this sort of things with aliases; there's an example in the [ES Index Aliases API](http://www.elasticsearch.org/guide/reference/api/admin-indices-aliases.html).

karmi:
> (I hope it's okay for @karmi to use this issue as a kind of support thread... :-))

Absolutely!! :) (I don't have time at the moment, but will return to this discussion later today or tommorrow.)

> blows the dust off...

:D Nice :)

rathgar:
Many thanks for the input, I've spent a fair bit of time digging and patch testing but still can't seem to achieve this cleanly. 

I set up the `indices` and `aliases` by hand so they are in place.
**/_aliases/** returns: `{ "s4-events" : { "aliases" : { "events" : { } } }, "s3-events" : { "aliases" : { "events" : { } } } }`
It works exactly as you'd expect for search. I also hoped that Tire would be able to update the document via the alias but this is not the case. Tire doesn't update the index and some digging around with `curl` reveals what es thinks of this:
`{ "error" : "ElasticSearchIllegalArgumentException[Alias [events] has more than one indices
     associated with it [[s4-events, s3-events]], can't execute a single index op]",
   "status" : 400 }` 
which, come to think of it, is perfectly understandable - you could easily have the same document in multiple indices.

This means that es needs to be given the actual `index` name for create, update & delete actions on the index. Search of course is the easy bit, you can pass that in within custom search wrapper methods on the models.

There are two routes I can think of:
1. write my own callbacks to handle the (re)indexing;
2. get Tire to understand how the indices are named and let it do all the heavy lifting.

**Callbacks**

On the face of it, this seems  the easy route. However, seeing as es needs to know the name of the `index`, it looks like you would need to rewrite Tire's `update_index` to get this to work. Replicating good, working code never feels right.

**index_name**

The smoothest solution I can think of is if models could define an instance level `index_name`. This would override the default instance method defined in `lib/tire/model/naming.rb` which is just a proxy to the class level method. So far so good.

From what I can tell Tire always calls the class level version of this method. If it were to always call the instance level method this would result in the same response to all requests unless the model overrode it, resulting in custom index name.

I'm not sure if doing this would conflict with what @karmi has in mind for Tire, I'm open to direction and/or smack-down `:-P`

olivere:
Yes, searching should be fine.

Regarding the indexing, you cannot index against an alias (as you found out). I'd go for: Disable the Tire callbacks and manage the indexing yourself, sort of like shown in the README. Shouldn't be too hard. That'd have least impact and maximizes your chances it will work while Tire gets updated. Maybe you can find an example in the integration tests, or @karmi can give you a hint as to what is a future-proof way of doing so.

karmi:
Hi, thanks to @olivere for the input and great advice, couldn't say it better. I'd like to add a couple of points:
- As Oliver said, elasticsearch does not care if you pass an index or alias to many operations, such as searching -- it automatically expands it. Note that on 0.19.8 there's even a wildcard notation where you can say `products-*`.
- You actually can index against an alias, but it has to point to a _single_ index -- otherwise elasticsearch wouldn't know into which of the aliased index to put the data, obviously.
- The whole story with “dynamic” or “calculated” index name is still too confusing to me. I have tried to re-read the thread and apart from great memories didn't get much from it :)

Notice that _Tire_ has support for the “dynamic index name” as in “lambda that is evaluated in the context of the class”, see the [test model](https://github.com/karmi/tire/blob/master/test/models/active_record_models.rb#L72-79) or the [relevant code](https://github.com/karmi/tire/blob/master/lib/tire/model/search.rb#L100-109).

If I remember correctly, sketches and spikes involving index names bound to the instance context were leading nowhere interesting. More importantly, I think this is precisely the spot where it could be wise to deviate off the “just let the library handle that” course and implement your own solution on top of the library.

In the applications I have been working on, there has been a similar need for separating documents into indices “per user” or “per account”. I can summarize here what we did, in two separate scenarios.

Let's say we have an `Account` class and we deal in `articles` entities.

In that case, our `Account` class would have following:

``` ruby
class Account
  #...

  # Set index name based on account ID
  #
  def articles
      Article.index_name "account-#{self.id}"
      Article
  end
end
```

So, whenever we need to access articles for a particular account, either for searching or for indexing, we can simply do:

``` ruby
@account = Account.find( remember_token_or_something_like_that )

# Instead of `Article.search(...)`:
@account.articles.search { query { string 'something interesting' } }

# Instead of `Article.create(...)`:
@account.articles.create id: 'abc123', title: 'Another interesting article!', ...
```

**BOOM!** Done :)

Having a separate index per user/account works perfect in certain cases -- but definitely not well in cases where you'd have tens or hundreds of thousands of indices (or more). Having index aliases, with properly set up filters and routing, would perform much better in this case. We would slice the data not based on the tenant identity, but based on time.

Let's have a look at a second scenario, starting with a heavily simplified `curl http://localhost:9200/_aliases?pretty` output:

``` json
{
  "articles_2012-07-02" : {
    "aliases" : {
      "articles_plan_pro" : {
      }
    }
  },
  "articles_2012-07-09" : {
    "aliases" : {
      "articles_current" : {
      },
      "articles_shared" : {
      },
      "articles_plan_basic" : {
      },
      "articles_plan_pro" : {
      }
    }
  },
  "articles_2012-07-16" : {
    "aliases" : {
    }
  }
}
```

You can see that we have three indices, one per week. You can see there are two similar aliases: `articles_plan_pro` and `articles_plan_basic` -- obviously, accounts with the “pro”  subscription can search two weeks back, but accounts with the “basic” subscription can search only this week.

Notice also, that the the `articles_current` alias points to, ehm, current week (I'm writing this on Thu 2012-07-12). The index for next week is just there, laying and waiting -- when the time comes, a background job (cron, _Resque_ worker, custom script, ...) will update the aliases. There's a nifty example with aliases in “sliding window” scenario in the _Tire_ [integration test](https://github.com/karmi/tire/blob/master/test/integration/index_aliases_test.rb#L66-120) suite.

Let's not look on the `articles_shared` alias right now, let's look at what tricks we can play with this setup:

``` ruby
class Account
  # ...

  # Set index name based on account subscription
  #
  def articles
    if plan_code = self.subscription && self.subscription.plan_code
      Article.index_name "articles_plan_#{plan_code}"
    else
      Article.index_name "articles_shared"
    end
    return Article
  end
end
```

Again, we're setting up an `index_name` for the `Article` class, which holds our documents. When the current account has a valid subscription, we get the `plan_code` out of the subscription, and direct searches for this account into relevant index: “basic” or “pro”.

If the account has no subscription -- he's probably a “visitor” type -- , we direct the searches to the `articles_shared` alias. Using the interface is as simple as previously, eg. in `ArticlesController`:

``` ruby
@account  = Account.find( remember_token_or_something_like_that )
@articles = @account.articles.search { query { ... } }
# ...
```

We are _not_ using the `Article` class as a gateway for indexing in this case; we have a separate indexing component, a _Sinatra_ application serving as a light proxy to elasticsearch _Bulk API_, providing HTTP authentication, document validation (enforcing rules such as required properties or dates passed as UTC), and uses the bare `Tire::Index#import` and `Tire::Index#store` APIs.

These APIs talk to the `articles_current`index alias, which is periodically updated to the current week with said background process. In this way, we have decoupled all the logic for setting up index names in separate components of the application, so we don't need access to the `Article` or `Account` classes in the indexing proxy (it runs on a separate server), or any component of the application. _Whichever_ component is indexing, indexes against `articles_current` alias; _whichever_ component is searching, searches against whatever alias or index makes sense for the particular component.

As Oliver rightly points out, _not_ using _Tire_ for indexing is perfectly valid and very simple: just do _not_ `include` the `Tire::Callbacks` module. (Side note: Based on the feedback I get, many people expect _Tire_ to magically guess and support their needs; I think, and have repeatedly said, that in many cases, you just _have_ to sit down and write your thing. In my experience, due to _Ruby_ being so awesome and _elasticsearch_ being so awesome, _Tire_ can support all sorts of tricks like these.)

For anybody dealing with complex scenarios like these, I strongly suggest to watch @kimchy's _Berlin Buzzwords 2012_ talk [“ElasticSearch: Big Data, Search, and Analytics”](http://vimeo.com/album/1968418/video/44716955) ([slides](https://speakerdeck.com/u/kimchy/p/elasticsearch-big-data-search-analytics)) -- it walks through all sorts of different scenarios and topologies for big(gish) datasets.

olivere:
Thanks for your insights, @karmi. Really appreciate that.

IIRC we did something similar to your `articles` method in the `Account` class... something along the lines of:

``` ruby
class Product
  # ...

  index_name { Thread.current[:product_index_name] }

  def self.with_index(name, &block)
    old_name = Thread.current[:product_index_name]
    Thread.current[:product_index_name] = name
    yield block
    Thread.current[:product_index_name] = old_name
  end
end
```

... which leads to ...

``` ruby
Product.with_index(index_name) do
  # ...
end
```

If it works, I claim copyright... Otherwise, well... ;-)

karmi:
@olivere That looks as a great solution, actually! I think that's something which could be supported in _Tire_ quite well. What's the motivation for the multi-threaded support, though? :)

olivere:
Copyright, yay! ;-)

Maybe its over-engineered, but I'd like to be on the safe side. This would work fine even in a multi-threaded web server with JRuby, wouldn't it?

rmm5t:
+1 to @olivere's approach. Given the current Tire API, this looks to be the most elegant solution that both supports multi-tenant indexes and is thread-safe.

farukca:
@olivere I have customer model which is seperated with current_account_id; and my customer model's index name like;

``` ruby
index_name { "customers-#{Thread.current[:current_account_id]}" }
```

When I add new record it is working perfect, but how can I populate the index for existing records ?

Thanks, best regards...

olivere:
@farukca The idea is basically to wrap access to Elasticsearch indices and make sure that `Thread.current[:current_account_id]` is set before `index_name` is evaluated. One way of doing this is to use it as outlined in https://github.com/karmi/tire/pull/92#issuecomment-6930576.

I generally raise an error to ensure I didn't forget to set the current account, like so:

```
index_name { Thread.current[:current_account_id] or raise "Wrap search with Customer.with_index(...)" }
```

The stack trace might give you an idea where you forgot to add the `with_index` wrapper.

BTW: You can of course always use `Tire.search(...)` and friends directly and provide the index name (manually).

Hope that helped...

PR #97, opened 2011-09-07
Support for an index prefix for easy multi-environment support (test, development, production etc.)

redbeard:
This change is a simple solution, similar to couchrest_model's approach, that allows setting a prefix for all automatically generated index names. 

This setting _only_ affects default index names that have not been overridden by a custom index name.

Furthermore, the default behaviour is no prefix, so it will not affect existing codebases.

One potential usage is a per-environment prefix, configured like so:
`Tire.configure { index_prefix "#{Rails.env.to_s.downcase}_" }`

This will result in a 'test_articles' index for Article instances when used in tests, and 'development_articles' when used in development.

This change includes:
- Tests
- Tiny amount of code changes 
- Usage info in README.markdown 

I hope you accept it into master :)

Cheers,
Tal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ scratch/
 examples/*.html
 *.log
 .rvmrc
+.redcar
--- a/README.markdown
+++ b/README.markdown
@@ -356,7 +356,7 @@ When you now save a record:
                    :published_on => Time.now
 ```
 
-it is automatically added into the index, because of the included callbacks.
+it is automatically added into an index called 'articles', because of the included callbacks.
 (You may want to skip them in special cases, like when your records are indexed via some external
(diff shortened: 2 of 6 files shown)


karmi:
Hi, thanks for the suggestion! I understand and like the idea. I was playing with it a bit today, and think it would be better to expose something like that just for the models; `Tire.config` seems a bad location for a feature like that?

I think we should expose it via a model class method, which you could use either on a model-per-model basis, or generally, in an initializer or common config file, a la `Tire::Model::Search.index_prefix 'myprefix'`... I'll definitely put some time to it!

flockonus:
Awesome feature!

Suggestion for `Tire::Model::Search.index_prefix 'test'` to go to docs, super handy for tests

Issue #60, opened 2011-07-16
Make Tire HTTP-Client Agnostic


karmi:
Currently, _Tire_ is closely coupled with the [_RestClient_](https://github.com/archiloque/rest-client) HTTP library on the implementation level, although it is prepared for pluggable backends on the architecture level.

To make this architectural support, such as `Tire.configure { client MyClient }`, really useful, and HTTP backend pluggable, the following must be implemented:
- Create a wrapper class `Tire::HTTP::Response` for HTTP responses, with uniform interface for getting response body, code, headers.
- Use the wrapper class in the default client implementation, `Tire::Client::RestClient`
- Use the wrapper class throughout the code (eg. in `Tire::Search::Search.perform` method)
- Create another client implementation for testing purposes, preferably for [_Curb_](https://github.com/taf2/curb)

The goal is to make the implementation of another HTTP client as easy as providing an interface in _Tire_ or [_Tire Contrib_](https://github.com/karmi/tire-contrib).

selvakn:
Hi Karel

I started working on this to make use of em-http-client based client. Then noticed most of the api methods return false if the operation fails , but some (ex: bulk_update) does raise RestClient exceptions back. Im trying to pull out all the references to RestClient out. So does it make sense to make all the methods to return false on failure?

karmi:
OK, it should be easy now to add more HTTP clients, there's a _Curb_ one in the repo.

Based on some very basic benchmarks [https://gist.github.com/1204159], Curb seems to be more then two times faster then RestClient, even though the results do fluctuate a lot.

I'm wondering about the em-http implementation? How would the async approach here?

PR #98, opened 2011-09-07
Issue #60 (Make Tire HTTP-Client Agnostic)

selvakn:
Implemented Tire::HTTP::Response, Made default implementation(RestClient) to use Tire::HTTP::Response, All api calls returns false on failure.


karmi:
Thanks!, pulled in and cleaned up. I still don't know if the energy wouldn't be better directed at making _Tire_ use [_Faraday_](https://github.com/technoweenie/faraday), but... this is clearly better then it was.

PR #99, opened 2011-09-08
Support for an index prefix for easy multi-environment support (test, development, production etc.) based in Model::Search

redbeard:
Based on your feedback, reimplemented karmi/tire#97 to use Model::Search setting rather than Tire::Configuration.

This change is a simple solution, similar to couchrest_model's approach, that allows setting a prefix for all automatically generated index names. 

This setting _only_ affects default index names that have not been overridden by a custom index name.

Furthermore, the default behaviour is no prefix, so it will not affect existing codebases.

One potential usage is a per-environment prefix, configured like so:
`Tire::Model::Search.index_prefix "#{Rails.env.to_s.downcase}_"`

This will result in a 'test_articles' index for Article instances when used in tests, and 'development_articles' when used in development.

This change includes:
- Tests
- Tiny amount of code changes 
- Usage info in README.markdown 

Cheers,
Tal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ scratch/
 examples/*.html
 *.log
 .rvmrc
+.redcar
--- a/README.markdown
+++ b/README.markdown
@@ -693,7 +693,7 @@ You can do so by configuring an index_prefix that will be used to prefix the def
 Add to your initializers a similar snippet to the following:
 
 ```ruby
-    Model::Search.index_prefix "#{Rails.env.to_s.downcase}_"
+    Tire::Model::Search.index_prefix "#{Rails.env.to_s.downcase}_"
 ```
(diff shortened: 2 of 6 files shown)


karmi:
Hi Tal, thanks! I've merged your commit and then refactored it a bit. It now allows for setting the `index_prefix` either for all models, or for specific models only -- I've felt it has more balance this way, even though the implementation (class variables, checking context, ...) is far from ideal.

Could you test it in your app, if it works as expected?

redbeard:
Thanks for merging!

Your refactoring works well in my app as far as the global models prefix is concerned.
I haven't tried the per-model prefix as of yet.

As you're so responsive, I expect more patches to be forthcoming as I continue using Tire :)

PR #103, opened 2011-09-15
Instrumentation

olivere:
This pull request includes basic support for instrumenting Tire in a Rails app.

It basically adds search runtime to the Rails log, like this:

```
Started GET "/articles?utf8=%E2%9C%93&q=bull*&commit=search" for 127.0.0.1 at 2011-09-15 19:07:00 +0200
  Processing by ArticlesController#index as HTML
  Parameters: {"utf8"=>"✓", "q"=>"bull*", "commit"=>"search"}
  Tire search (6.7ms)  {"query":{"query_string":{"query":"bull*"}}}
  Article Load (0.3ms)  SELECT `articles`.* FROM `articles` WHERE `articles`.`id` IN (104126575, 146022759)
Rendered articles/index.html.erb within layouts/application (13.0ms)
Completed 200 OK in 63ms (Views: 53.7ms | ActiveRecord: 1.8ms | Tire: 6.7ms)
```

Let me know what you think.

### DIFF
--- a/lib/tire.rb
+++ b/lib/tire.rb
@@ -40,3 +40,6 @@ module Tire
   end
   module_function :warn
 end
+
+require 'tire/rails/railtie' if defined?(Rails)
+
--- a/lib/tire/rails/controller_runtime.rb
+++ b/lib/tire/rails/controller_runtime.rb
@@ -1,0 +1,34 @@
+require 'active_support/core_ext/module/attr_internal'
+
+module Tire
+  module Rails
+    module ControllerRuntime
+      extend ActiveSupport::Concern
+
+    protected
+
+      attr_internal :tire_runtime
+
+      def cleanup_view_runtime
+        tire_rt_before_render = Tire::Rails::LogSubscriber.reset_runtime
+        runtime = super
+        tire_rt_after_render = Tire::Rails::LogSubscriber.reset_runtime
+        self.tire_runtime = tire_rt_before_render + tire_rt_after_render
+        runtime - tire_rt_after_render
+      end
+
+      def append_info_to_payload(payload)
+        super
+        payload[:tire_runtime] = (tire_runtime || 0) + Tire::Rails::LogSubscriber.reset_runtime
+      end
+
+      module ClassMethods
+        def log_process_action(payload)
+          messages, tire_runtime = super, payload[:tire_runtime]
+          messages << ("Tire: %.1fms" % tire_runtime.to_f) if tire_runtime
+          messages
+        end
+      end
+    end
+  end
+end
--- a/lib/tire/rails/instrumentation.rb
+++ b/lib/tire/rails/instrumentation.rb
@@ -1,0 +1,20 @@
+module Tire
+  module Rails
+    module Instrumentation
+      extend ActiveSupport::Concern
+
+      included do
+        alias_method_chain :perform, :instrumentation
+      end
+
+      module InstanceMethods
+        def perform_with_instrumentation
+          # Wrapper around the Search.perform method that logs search times.
+          ActiveSupport::Notifications.instrument("search.tire", :name => 'Tire search', :search => self.to_json) do
+            perform_without_instrumentation
+          end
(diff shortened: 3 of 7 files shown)


karmi:
Hi, this is a great addition. The implementation looks a bit complicated, though? The ones I know from the Jose Valim's book and eg. http://asciicasts.com/episodes/249-notifications-in-rails-3 are much more lightweight?

@joshk: Any feedback, please?

olivere:
Sure, will have a look at simplifications (still on my journey to Ruby-zen). The implementation has been "borrowed" from several other Rails plugins and my own understanding of the Notifications API.

Some questions still:

a) The log subscriber integration test copies a lot from the AR tests (schema setup etc.). Do you prefer to separate test files on the scenario (Notifications,  LogSubscriber etc.), or do you accept adding it to the AR tests?

b) I wanted to keep Tire unaware of Rails, therefore only introduce the Railtie and the rest of the instrumentation in a Rails environment. As a result, I wrapped the perform method in a method chain, which is part of what looks complicated (I guess). I could simplify this and get rid of the method chain, but that probably means that the Notification API gets into Tire itself. Any thoughts on this?

c) This patch applies to Rails 3.1. Still have to test it with 3.0 and 2.x.

olivere:
Just looked up the sample in Jose Valims book and on Railscasts: They're more lightweight because they focus on consuming events. They don't append e.g. the search runtime to the Rails action log (as shown [here](https://gist.github.com/566725)). [`LogSubscriber`](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/log_subscriber.rb) and [`ControllerRuntime`](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/railties/controller_runtime.rb) are basically copied from ActiveRecord. Looking at other gems, I found this to be pretty standard, but I might be wrong. I couldn't find an "official" guide for this. Do you have any experts you might ask?

karmi:
Hi,

yeah, achieving what you were after is more complicated then I thought.

I have cheated :), and extracted your commit as a first component for the [`karmi/tire-contrib`](https://github.com/karmi/tire-contrib) gem; https://github.com/karmi/tire-contrib/commit/b900c29b5e5c778c3a18ebc73a9585ffb5710fcc

Is that OK?

I envision `tire-contrib` as a place for all sorts of additions and extensions to _Tire_, in a spirit similar to [`rack-contrib`](https://github.com/rack/rack-contrib/).

olivere:
Sure. Makes sense... I have all sorts of ideas/features that are probably wrong in tire(-core).

karmi:
Yes, thanks for the feature!

jlecour:
Hi,

Being in the process of instrumenting some code that use Tire, I was looking for some internal instrumentation and found this PR.

`ActiveSupport::Instrumentation` is a really good basic framework for instrumentation, but you have to depend on ActiveSupport to use it. It might be quite overkill if you only want this and not the whole ActiveSupport code base.

Looking at the gemspec, ActiveSupport is not a dependency of Tire, but in the `lib/tire.rb` file there are some `require` of parts of ActiveSupport. so I guess Tire _does_ depend on ActiveSupport. Using `ActiveSupport::Instrumentation` is suddenly no longer a "no go".
1. it might be wise to explicitly state the dependency in the gemspec
2. what to you think of putting a simplified version (no _alias method chain_) of the instrumentation directly into Tire ?

Thanks

karmi:
Hi,

the instrumentation for Rails is extracted into `tire-contrib`: https://github.com/karmi/tire-contrib/blob/master/lib/tire/rails/logger.rb

ActiveSupport _is_ ultimately a dependency via the ActiveModel dependency. I agree we should do a better job of admitting it in the gemspec, I just haven't found enough time to test version conflicts etc.

> what to you think of putting a simplified version (no alias method chain) of the instrumentation directly into Tire

What kind of instrumentation and for what purpose? In any case, I think the best place for anything like that is `tire-contrib`.

PR #104, opened 2011-09-15
Fixing an issue in Rails 3 related to the :load helper and differences between how AR.find() behaves vs. AR.where()

szehnder:
Fixing an issue where the find method would break when you pass it an array of ids (that were retrieved from a elasticsearch query) that were a super-set of the ids in the project database.

In other words, Imagine an elastic search index for an entity called a User.

User.search('foo', :load => true).collect(&:id)
["1", "2", "3", "4"]

User.search('foo', :load => true).results
Error -- if your rails database only has ["1", "3"]

---

this is because of a difference in how find() responds to a super-set array compared with the where() ActiveRecord method.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -176,7 +176,7 @@ module Tire
         # for the search method.
         #
         def load(options=nil)
-          options ? self.class.find(self.id, options) : self.class.find(self.id)
+          options ? self.class.where(:id => self.id).where(options) : self.class.where(:id => self.id)
         end
 
       end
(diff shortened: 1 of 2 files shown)


karmi:
Hi,

I'd like to merge it, but it depends on `Model.where` method, which is specific to ActiveRecord. ActiveModel does not usually have such method.

I'm curious, how a situation like this could happen? Isn't the real issue that ES and the db get out of sync? If so, does it make sense to call `MyModel.tire.index.refresh`?

karmi:
Hi, any news on this?

Issue #114, opened 2011-09-29
Allow Manually Controlling Callbacks


Rio517:
Hi, 

I have a lot of ActiveModel documents I don't want indexed - almost 80% of my records.  I tried removing the `include Tire::Model::Callbacks`, but then get an undefined method error: 

<pre><code>`undefined method `_run_update_elasticsearch_index_callbacks' for ...</code></pre>

 from this line https://github.com/karmi/tire/blob/master/lib/tire/model/search.rb#L128

Is there some way around this?  How can I disable callbacks and take control over when update_index is called?  Otherwise, is there some way to filter records that are getting indexed?  For instance, I only want articles with a state of published to be indexed.

Thanks!

-Mario

karmi:
Hi, when do you get the exception? Because simply not including `Callbacks` should work. That's the initial reason for splitting the functionality. I cannot replicate the issue.

> I only want articles with a state of published to be indexed

No problem -- you have to solve it in your app, since it and only it knows the domain in that case. Something like:

``` ruby
    class Article

      after_save do
        self.update_index if self.state == 'published'
      end

      # ...

    end
```

Rio517:
Yep.  That is exactly the behavior I was expecting and exactly what I wanted to do.

Here is a sample app. Run rake db:migrate then  db:seed and it will create data, index it and run the command to generate the error.   Console output of my error is in the readme.

https://github.com/Rio517/tire-sample-error

-Mario

karmi:
Hi, thanks for the sample app, I figured out what's the problem. Please check on current master, should be fixed.

karmi:
Ping, @Rio517. Any success with the fix?

Rio517:
All working well! Thanks!

karmi:
Pefect!

PR #115, opened 2011-09-29
allow removal from index on update_index callback 

jarosan:
I'm working on a project where we have a much larger set of data than I would like to have in the index. For this reason I am using a should_be_indexed? method and if false then submit an empty json in to_indexed_json method.

When an object gets changed and should_be_indexed? conditions change and return false then I want to remove object from the index.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -126,7 +126,7 @@ module Tire
         #
         def update_index
           instance.send :_run_update_elasticsearch_index_callbacks do
-            if instance.destroyed?
+            if instance.destroyed? || !instance.should_be_indexed?
               index.remove instance
             else
               response  = index.store( instance, {:percolate => percolator} )
@@ -157,6 +157,10 @@ module Tire
             reject { |key, value| ! instance.class.tire.mapping.keys.map(&:to_s).include?(key.to_s) }.
             to_json
           end
+        end
+        
+        def should_be_indexed?
+          true
         end
 
         def matches
--- a/test/integration/active_record_searchable_test.rb
+++ b/test/integration/active_record_searchable_test.rb
@@ -74,6 +74,27 @@ module Tire
         assert_not_nil results.first._score
         assert_equal   'Test', results.first.title
       end
+      
+      should "remove document from index on save if should not be indexed" do
+        a = ActiveRecordArticle.new :title => 'Test'
+        a.save!
+        id = a.id
+
+        a.index.refresh
+
+        results = ActiveRecordArticle.search 'test'
+
+        assert       results.any?
+        assert_equal 1, results.count
+
+        a.update_attribute(:title, "should_not_be_indexed")
+        assert ! a.should_be_indexed?
+
+        a.index.refresh
+
+        results = ActiveRecordArticle.search 'should_not_be_indexed'
+        assert results.empty?
+      end
 
       context "with eager loading" do
         setup do
--- a/test/models/active_record_models.rb
+++ b/test/models/active_record_models.rb
@@ -32,6 +32,10 @@ class ActiveRecordArticle < ActiveRecord::Base
       :stats        => stats.map    { |s| { :pageviews  => s.pageviews } }
     }.to_json
   end
+  
+  def should_be_indexed?
+    title != 'should_not_be_indexed'
+  end
 
   def length
     title.length
--- a/test/unit/model_search_test.rb
+++ b/test/unit/model_search_test.rb
@@ -243,6 +243,16 @@ module Tire
 
           @model.destroy
         end
+        
+        should "remove from index if should not be indexed" do
+          @model = ActiveModelArticleWithCallbacks.new
+          @model.stubs(:id).returns(1)
+
+          @model.expects(:should_be_indexed?).returns(false)
+          Tire::Index.any_instance.expects(:remove)
+          @model.update_index
+        end
+        
 
         should "store the record in index on :update_elasticsearch_index when saved" do
           @model = ActiveModelArticleWithCallbacks.new


karmi:
Hi, I feel sorry for such a fine patch. You can, in fact, achieve what you're after if you do _not_ include `Tire::Model::Callbacks` and implement the indexing hooks manually, a la:

``` ruby
class Article

  after_save do
    self.update_index if self.state == 'published'
  end

  # ...

end
```

See this [comment](https://github.com/karmi/tire/issues/114#issuecomment-2234996) for issue karmi/tire#114.

karmi:
Ping! Any news on this? Were you able to achieve the custom indexing with example I posted?

ralph:
While this is certainly possible to achieve (at least with the new version 0.3.4), I like the `should_be_indexed?` variant better, because you don't have to take care of the destroy/removal from the index. It's an interface everybody instantly understands, like the `to_indexed_json` method. Include the default modules for the default behavior, overwrite a method if you want custom behavior, simple as that.

But anyway, thanks for making this possible with version 0.3.4.

leehambley:
I agree with @ralph's position here, but only in the face of Tire being effectively a "Rails" interface to full text search. In Rails apps, as with most "web applications" I think it's quite normal that there's a bunch of things which shouldn't be searchable... and it's natural to expect something like `should_be_indexed?` – I expected at least that it would exist. (but we haven't had the need for me to explore that more thoroughly yet)

karmi:
@ralph, @leehambley: I understand your point. I don't think adding a top level thing like `should_be_indexed?` is particularly useful to general audience, though.

First, you have to properly document and explain, how that works, and you'll still be fighting edge-cases. It's a slippery slope of massive "optional" behaviours written for singular cases. We should be more concerned about adding `timeout` and `search_type`, folks! :)

Second, and more importantly, _Tire_ is just Ruby -- why not use it? I mean, `Tire::Model::Search::InstanceMethods::Searchupdate_index` is just couple of lines, doing nothing particularly advanced. If one has an edge case, where the indexing logic is somehow more complicated, she should _definitely_ write her own indexing hooks. Possibly doing stuff in the background, via queue or messaging infrastructure, etc. 

All that said, I'll happily pull that into karmi/tire-contrib!

@jarosan: Any luck using what I suggested?

jarosan:
@karmi: The reason I wanted this to be done with `should_be_indexed?` method is because i wanted to keep the existing destroy/save callbacks without worrying about them. 

It is possible to achieve that with this:
```before_update_elasticsearch_index :remove_from_index
def remove_from_index
Place.tire.index.remove self if self.published_at.blank?
end

```

but this still felt like a hack to get it working. 

I think the reason my suggestion makes less sense to you because I didn't include the logic of indexing/not indexing the record in to_indexed_json method. @karmi, what do you think. Is it possible that you change your mind, if i submit a patch, where this logic (should_be_indexed?) would also affect whether an object gets indexed or not in the first place. That way this would be a general and recommended way to decide what objects get indexed and what not. 

The syntax for doing this could be different. Either a should_be_indexed? method. Or maybe use named_scope-like syntax to pass a proc `exclude_from_index lambda { |r| r.published_at.blank? }`

Either way I really think that this is something Tire would benefit from :)
```

**Comment by karmi** [comment]
Hi @jarosan,

I am still not convinced conditional indexing like this is something which belongs to core _Tire_. I still think you'd be better off if you took care of the indexing yourself, in this case. I imagine the tests, the code readability, everything would benefit from that.

That said, why not extract that, and put it into `tire-contrib`? Then the only thing you have to do is:

```
gem 'tire-contrib'
require 'tire/conditional_indexing'
```

I will happily pull that into [karmi/tire-contrib](https://github.com/karmi/tire-contrib).

**Comment by karmi** [comment]
@jarosan: Ping -- need any help with porting it over to `tire-contrib`?

**Comment by jarosan** [comment]
@karmi: no worries. wouldn't want to overcomplicate my code with using `tire-contrib`. will use the `before_update_elasticsearch_index` hook to get it working.

**Comment by mahemoff** [comment]
For anyone searching, I explained how to support private resources with manual callbacks. http://softwareas.com/private-resources-with-elasticsearch-and-tire

**Comment by karmi** [comment]
Thanks for the article, Michael, will have a look.

**Comment by brupm** [comment]
@karmi and @mahemoff 

Really confused about this: 

``` ruby
# This is copied straight from the source, which references half-baked implentations.
# Don't know in what situations it would be necessary, but including it anyway
before_destroy { @destroyed = true }
def destroyed?; !!@destroyed; end 
```

mahemoff:
@brupm it's code in the original callback [1] that implements the bult-in `destroyed?` method, but only @karmi can explain in what circumstances that method mightn't work.
1. https://github.com/karmi/tire/blob/master/lib/tire/model/callbacks.rb#L28

mahemoff:
@karmi I've just been looking at this code and it seems the `before_destroy` (or `destroyed?`) override is preventing destruction for some reason. Do you know if it's still necessary in Rails 3.2+?

(I wasn't much deleting this record type before, so it may have always been an issue.)

karmi:
@mahemoff As far as I remember, "some half-baked ActiveModel implementations" were not providing `destroyed?`, which would break the "update or delete" semantics in `update_index`. Is the issue you're experiencing on ActiveRecord, some ActiveModel, or Tire::Persistence?

mahemoff:
@karmi This is with an ActiveRecord where I'm manually implementing Tire::Model::Callbacks because some records are private. I noticed the records weren't being deleted until I removed those lines (Rails 3.2.13).

Based on your explanation, it's probably okay for me to remove those lines as I'm handling the callbacks manually anyway.

karmi:
@mahemoff This sounds like a bug -- would it be feasible to create an isolated example or even an integration test? No worries if not.

mahemoff:
I'll try to find time in the next few days.

I don't really think it's a Tire bug though.

PR #117, opened 2011-10-03
Error fixed when trying to highlight multi_field types.

cedufca:
When using multi_field types an error ocurred as ActiveRecord tried to set "description.partial" (for example) in the model class. This commit fixes it by using the first word in the field name for highlights. (which means it will still working with non multi_field types)

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -30,7 +30,10 @@ module Tire
 
                  # Update the document with meta information
                  ['_score', '_type', '_index', '_version', 'sort', 'highlight'].each { |key| document.update( {key => h[key]} || {} ) }
-
+                 
+                 highlight = document['highlight']
+                 # Removes unecessary (for highlight) extra names after the field name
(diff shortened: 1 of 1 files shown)


karmi:
Hi,

two things:

1/ Could you add unit tests for the fix?

2/ How does ActiveRecord relate to this? The object in `Tire::Collection` should be of type `Item`.

cedufca:
1) Sure.
2) Hmm I think I forgot Tire works without Rails. In this case, ActiveRecord was trying to set "description.partial" on my model class, when it must set "description" only. This made me think now: is there a way of having (by ElasticSearch) different highlights on the same field? If so, it may be better just replacing non-alphanumeric with an underline. "description.partial" would be "description_partial". What you think about it?

karmi:
>  ActiveRecord was trying to set "description.partial" on my model class

Yeah, and that's exactly the part where I am wondering how ActiveRecord can do that? :) Because from v0.2 on Tire returns only `Item` instances, and allows you to load the _real_, ActiveRecord models with the `load` method or the `:load` option to the `search()` method.

karmi:
Ping, any news?

PR #119, opened 2011-10-05
Fix rake task: apply settings on index creation and fix progress bar

rubish:
1. Settings are not applied if index is created from rake task. Added settings to index creation in rake task.
2. When using rake task with mongoid, progress bar goes over 100%. This is because mongodb ignores limit and skip parameters for a count query. Additional parameter must be passed to take limit into account. Ref: http://www.mongodb.org/display/DOCS/Advanced+Queries#AdvancedQueries-%7B%7Bcount%28%29%7D%7D


karmi:
Hi, thanks for the catch, the `settings` fix pushed in karmi/tire@28989ec.

As for the Mongobar, could you please extract that into a [**karmi/tire-contrib**](http://github.com/karmi/tire-contrib) extension?

I really don't want the Tire codebase to be sprinkled all over with if/defined/whatever Mongo just because all the Mongo gems and ecocystem play by their own rules.

rubish:
Will look into the **tire-contrib** gem and see if I can move the progress bar fix there.

PR #121, opened 2011-10-08
allow multiple calls to filter, and the filters defined by calls

rubish:
It seems there can be only one call to `filter` in a search block. If there are more, only latest filter will be used. It was not intuitive to me. I think if there are more than one calls to filter all the filters should be joined with **and**. Had done some changes to do the same.

Let me know if it is not what you want or if there can be a better way to do it.

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -89,7 +89,13 @@ module Tire
         request.update( { :query  => @query.to_hash } )    if @query
         request.update( { :sort   => @sort.to_ary   } )    if @sort
         request.update( { :facets => @facets.to_hash } )   if @facets
-        @filters.each { |filter| request.update( { :filter => filter.to_hash } ) } if @filters
+        if @filters
+          filters = []
+          @filters.each do |filter|
+            filters << filter.to_hash
+          end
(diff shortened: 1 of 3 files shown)


rubish:
I messed up something in the commit. Haven't worked with git before. I only wanted to add pull request for d6eb52efc9413eae76ab016fd855556b2922756b and b1283edc0a717f4769c53c60f5524d1340c5dfcb.

karmi:
I like the direction, and yeah, we should make it possibly to just call `filter` multiple times to add more filters. However, the problem in your code is you just add `and` all the time.

Eg. when you add a `puts s.to_json` in here: https://github.com/karmi/tire/blob/master/test/integration/filters_test.rb#L20-25, you get this:

```
"filter":{"and":[{"or":[{"terms":{"tags":["ruby"]}},{"terms":{"tags":["erlang"]}}]}]}
```

Not ideal.

rubish:
Would changing it to add `and` only if there are two or more filters suffice or you want it to do something more?

rubish:
I have created a new fork, to fix the messed up commits in my repo. Will resubmit with using `and` only if there are two or more filters present.

PR #122, opened 2011-10-11
Allow multiple filter calls to be joined with 'and'

rubish:
Added test cases to demonstrate partial support for multiple filters, but being ignored while generating search json. Added support to join multiple filters with `and`.

I guess eventually filters should have something like `Query#boolean`.

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -89,7 +89,8 @@ module Tire
         request.update( { :query  => @query.to_hash } )    if @query
         request.update( { :sort   => @sort.to_ary   } )    if @sort
         request.update( { :facets => @facets.to_hash } )   if @facets
-        @filters.each { |filter| request.update( { :filter => filter.to_hash } ) } if @filters
+        request.update( { :filter => @filters.first.to_hash } ) if @filters && @filters.size == 1
+        request.update( { :filter => { :and => @filters.inject([]) { |filters, filter| filters << filter.to_hash } } } ) if  @filters && @filters.size > 1
         request.update( { :highlight => @highlight.to_hash } ) if @highlight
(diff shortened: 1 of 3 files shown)


chrisberkhout:
This sounds good.

A side note: my current (v0.3.5) workaround for doing multiple filters is to do an `:and` filter and give it an array:

```
s = Tire.search("myindex") do
  filter "and", [
      {"or" => [
        { "term" => { "_type" => "document" } },
        { "term" => { "_type" => "source" } }
      ]},
      { "ids" => { "values" => ["4e8e8aa116e1771ef1000002"] } }
    ]
  query { all }
end
```

Which generates a call like:

```
curl -X GET "http://localhost:9200/myindex/_search?pretty=true" -d '
{
    "query": {
        "match_all": {}
    },
    "filter": {
        "and": [
            {"or": [
                { "term": { "_type": "document" }},
                { "term": { "_type": "source" }}
            ]},
            {"ids": {"values": ["4e8e8aa116e1771ef1000002"]}}
        ]
    }
}
'
```

Note: I previously thought the array had to be double wrapped (ie. `[[...]]`) for Tire to handle it correctly, but that was a mistake. A double wrapped array there is handled incorrectly by ES (following values are ignored).

rubish:
@chrisberkhaout: Yes you can certainly do that, this would allow one alternative of doing the same:

``` ruby
s = Tire.search("myindex") do
  filter "or", [
      { "term" => { "_type" => "document" } },
      { "term" => { "_type" => "source" } }
    ]
  filter "ids", "values" => ["4e8e8aa116e1771ef1000002"]
  query { all }
end
```

karmi:
Thanks to the original implementation, @rubish, code cleaned, pushed, released.

PR #123, opened 2011-10-12
Allow to create mapping for type nested

rubish:
There is currently no way to define [nested type](http://www.elasticsearch.org/guide/reference/mapping/nested-type.html) in mapping.. This pull request adds support for defining nested mapping type by passing options and adds tests around it.


karmi:
Removed the integration tests (really not desired for a edge feature like this), cleaned, pushed, released.

PR #124, opened 2011-10-12
add support for analyze api

rubish:
Adds support for [analyze api](http://www.elasticsearch.org/guide/reference/api/admin-indices-analyze.html). Very helpful when debugging why search is not working as expected, what terms a particular analyzer will generate or what terms ES is generating for the field.

Usage:

``` ruby
index.analyze("karmi/tire")
index.analyze("karmi/tire", :analyzer => 'whitespace')
index.analyze("karmi/tire", :field => 'title')
```

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -181,6 +181,19 @@ module Tire
       logged(error, '_close', curl)
     end
 
+    def analyze(text, options={})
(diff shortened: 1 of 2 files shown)


karmi:
Added more tests, cleaned up the Ruby code, pushed, released.

PR #127, opened 2011-10-14
Dynamic temples in a model, bulk load sleep on retries, and 

sakrafd:
This adds the ability to define dynamic temples in a model's mapping, configure a sleep time between retries for bulk load, and allows setting fields in query string search.

Thanks,
Dave

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -87,7 +87,13 @@ module Tire
       rescue Exception => error
         if count < tries
           count += 1
-          STDERR.puts "[ERROR] #{error.message}, retrying (#{count})..."
+          if ENV['BULK_RETRY_SLEEP_SECS']
+            sleep_time = count * ENV['BULK_RETRY_SLEEP_SECS'].to_i
+            STDERR.puts "[ERROR] #{error.message}, sleeping for #{sleep_time}sec & retrying (#{count})..."
+            sleep(sleep_time)
+          else
+            STDERR.puts "[ERROR] #{error.message}, retrying (#{count})..."
+          end
           retry
         else
           STDERR.puts "[ERROR] Too many exceptions occured, giving up. The HTTP response was: #{error.message}"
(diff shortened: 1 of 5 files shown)


vhyza:
Hello @sakrafd. I'd like to ask how many documents are you importing using bulk. I was using _bulk with about 1000 documents per request and I never noticed timeout. I had similar problems when ES was running on FreeBSD. Is there any chance to run import suite on different environment? Or try it with less documents per request. I'm just curious because I've never noticed this behaviour on OS X or linux.

sakrafd:
Hey @vhyza,

I tried batches of docs between 10K-1K all resulting in timeouts - sometimes the five retries that are already in tire would be enough to keep the process going, but it would eventually end up failing. Once I put in the sleep time, I settled on batches of 5k and was able to get through the remaining docs. We're on RHEL5 (and unfortunately can't change that in production), ES was at 17.6 when I imported the docs to a 3 server cluster with 10 shards and 1 replica.

Were you running batches of 1000 docs on FreeBSD when you were seeing timeouts? How many total docs did you import? How about mappings, did you have a lot of dynamic ones? Seems like there are a lot of variables that can contribute to the timeouts. I didn't see any timeouts on osx or our staging server which is also RHEL5, but it's not set up as a cluster and I wasn't importing nearly as many docs. My colleague is in the process of adding another index to our prod servers that will have 175M (docs without any dynamic mappings) to the same ES instance (on 17.8 now).  I'll let you know if he runs into timeouts.

dave

karmi:
10 shards is quite high number, IMHO. The cluster takes some heavy pounding in this case.

5000 docs per batch also seems like a big number to me...

Also, I suspect the real culprit is dynamic mapping, there is really an overhead with that. That's why I suggested we takes this on IRC, so we can pull kimchy into it.

karmi:
Hi, I've cleaned up a bit and merged the sakrafd/tire@c0c2c85 commit, thanks!

BTW, the proper way around this would be to eat all the valid options for _search endpoint, that would solve many issues such as `search_type=count`, etc. Will have to look into that.

PR #129, opened 2011-10-18
Require rdoc before sdoc, otherwise any rake operation fails

michaelklishin:


karmi:
Michael, I've removed SDoc completely, it seems to break on current RDoc and I have no time to investigate it.

PR #130, opened 2011-10-18
Added support for mapping fields

jwood:
I added support for the multiple fields (_source, _all, etc) that are supported by the elasticsearch mapping when the index is defined using the ActiveModel integration.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -91,6 +91,7 @@ module Tire
           retry
         else
           STDERR.puts "[ERROR] Too many exceptions occured, giving up. The HTTP response was: #{error.message}"
+          raise
         end
       ensure
         curl = %Q|curl -X POST "#{Configuration.url}/_bulk" -d '{... data omitted ...}'|
--- a/lib/tire/model/indexing.rb
+++ b/lib/tire/model/indexing.rb
@@ -35,7 +35,7 @@ module Tire
         #
         #     class Article
         #       # ...
-        #       mapping do
+        #       mapping :_source => { :compress => true } do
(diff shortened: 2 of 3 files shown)


karmi:
John, thanks for the patch. I'll look into that soon -- I'll just change couple of things, like adding a separate test, not changing existing, and maybe changing the name: `fields`.

I know it's consistent with the http://www.elasticsearch.org/guide/reference/mapping/boost-field.html naming, but I think `options` will be more consistent with general Ruby coding style and Tire codebase... What do you think?

jwood:
That sounds good to me.  Thanks!

karmi:
Hi John, I've cleaned up the patch a bit, reverted the change to test, and added a separate test for the feature (`should "create the index with proper mapping options"`).

Thanks!, K.

karmi:
Also, it's much more convenient for the administration, when heterogenous patches do not end up in one pull request (such as the "re-raise" one :) Usually, in Git, it's a very good idea to create a branch for every logical patch/feature...

jwood:
Yeah, sorry.  I didn't mean to complicate things with that other commit,
which I did not intend to be merged.  I forgot that the pull request was
still open.

Thanks again for merging the patch.

John

On Mon, Oct 24, 2011 at 3:14 AM, Karel Minarik <
reply@reply.github.com>wrote:

> Also, it's much more convenient for the administration, when heterogenous
> patches do not end up in one pull request (such as the "re-raise" one :)
> Usually, in Git, it's a very good idea to create a branch for every logical
> patch/feature...
> 
> ## 
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/karmi/tire/pull/130#issuecomment-2500767

## 

http://johnpwood.net

PR #132, opened 2011-10-24
Add support for range queries

ralph:


karmi:
Ralph, outstanding patch, thanks!

ralph:
Wheeeeeee!

PR #135, opened 2011-10-25
Using the wrapper option shouldn't change Configuration.wrapper.

dylanahsmith:
Fixes a TODO marked in the code.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -18,10 +18,12 @@ module Tire
 
       def results
         @results ||= begin
+          hits = @response['hits']['hits']
           unless @options[:load]
-            @response['hits']['hits'].map do |h|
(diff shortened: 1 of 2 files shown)


karmi:
Hi, just yesterday I was thinking the `:wrapper` option is redundant, fragile, and really not much usable anyway. In your implementation it does actually makes sense. Are you using it somehow, or was it just formal refactoring?

dylanahsmith:
I was just setting it to Hash to skip a bit of code when doing eager loading manually afterwards.  Nothing essential.

The only advantage to having the option it would be for multithreading, which isn't my use case.  For single threading Configuration.wrapper can be set before and after a request to do the same thing.

karmi:
Thanks, Dylan, merged & pushed.

PR #139, opened 2011-10-29
Added support for custom_score

### LINKED ISSUE #133: Add support for custom_score
Would be nice if there was a way to set the custom_score property for a query.  This allows for some pretty complex result ordering that is useful for a lot of applications.

pacoguzman:
Maybe this resolve #133

Excuse me but I don't know how to relate this pull request with the issue directly

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -28,6 +28,13 @@ module Tire
         @value
       end
 
+      def custom_score(options={}, &block)
+        @custom_score ||= Query.new(&block)
+        @value[:custom_score] = options
+        @value[:custom_score].update({:query => @custom_score.to_hash})
+        @value
+      end
+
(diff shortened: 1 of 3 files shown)


karmi:
Hi, terriffic patch, merged & pushed.

Also, I've added couple of _score computation "more real-word" examples to integration tests.

BTW: Adding commits to existing pulls is very easy. The best way is to create a branch for the issue (like you did), and then just keep pushing there with force. You can rebase, squash commits, change them, whatever, and then just push with `--force`. But it's a non-issue, practically, I can make sense of it quite easy on my end :)

BTW naming the branch `issues/133` has interesting benefits for you, such as "folder-like" display in GitX etc. :)

pacoguzman:
You're examples have more sense "more real-world" great!

Thanks for the notes

PR #142, opened 2011-11-03
on debug level, the logger will fail when performing a search fails

gtarnovan:
e.g. if the elastic search server is not running (or other causes if SearchRequestFailed in Tire::Search::Search#perform) the logger try to print the a null body and results in:

activesupport (3.1.0) lib/active_support/whiny_nil.rb:48:in `method_missing'
tire (0.3.8) lib/tire/logger.rb:50:in`log_response'
tire (0.3.8) lib/tire/search.rb:124:in `logged'
tire (0.3.8) lib/tire/search.rb:80:in`perform'
tire (0.3.8) lib/tire/model/search.rb:97:in `search'
tire (0.3.8) lib/tire/model/search.rb:264:in`**send**'
tire (0.3.8) lib/tire/model/search.rb:264:in `search'

### DIFF
--- a/lib/tire/logger.rb
+++ b/lib/tire/logger.rb
@@ -47,7 +47,7 @@ module Tire
       content += " [#{status}]"
       content += " (#{took} msec)" if took
       content += "\n#\n" unless json == ''
-      json.each_line { |line| content += "# #{line}" } unless json == ''
+      json.each_line { |line| content += "# #{line}" } unless (json == '' || json.nil?)
       content += "\n\n"
       write content
     end


karmi:
Hi, in the end, I ended up refactoring the mess for logging results in Index methods, and also added a "blank?" check in the location you were changing. Thanks for the report!

gtarnovan:
Hi! Thanks for the quick response.
Why don't you use ActiveSupport::BufferedLogger instead of rolling your own ? Or let the user supply one.

Thank you for your work!

karmi:
What do you mean by "let the user supply one"? You can pass an instance of any `IO`-compatible class to `Tire.configure { logger ... }`.

gtarnovan:
I mean, let the user supply a logger instance like Log4r, standard ruby logger, etc; just like Rails does with config.logger

karmi:
First, Tire's "logger" servers a different purpose then the usual Rails/ActiveRecord logger. It's purpose is to output a nicely formatted curl-based session transcript of talking to ES.

Second, Tire's logger does take any instance of any `IO`-based class (based on `respond_to?(:write)` check), so you're free to pass your custom objects. There hasn't really been a pressing and real need to explore it further.

There's a very nice ActiveSupport notification based Rails logger in contrib, see https://github.com/karmi/tire-contrib/blob/master/lib/tire/rails-logger.rb.

I was playing a bit with it, and made it a bit easier to pass Logger, Log4R instances to the `logger` method, so thanks for the hint!

karmi:
For an example:

``` ruby
require 'fileutils'
require 'active_support/buffered_logger'

mylog = ActiveSupport::BufferedLogger.new 'buffered-logger.log'
mylog.instance_eval do
  alias :write :info
end

require 'tire'

Tire.configure { logger mylog }

Tire.index('logger-articles') do
  delete
  create
  store :title => 'One', :tags => ['ruby']
  refresh
end

Tire.search('logger-articles') { query { string 'One' } }
```

PR #150, opened 2011-11-16
Add minitest as a development dependency in the gemspec

coffeencoke:


karmi:
Hi, why is this needed?

coffeencoke:
When running rake I got something like "you must install minitest, add minitest to your bundle. It's a development dependency. 

Thanks!

Matt

karmi:
Hi, weird, what Ruby version this is?

(AFAIK: Ruby 1.9.x should have Minitest wrapped in Test::Unit, which is the default on 1.8, so we're not forcing another specific gem on 1.8 users here. Cf. http://endofline.wordpress.com/2010/12/10/ruby-standard-library-mini-test/.)

karmi:
Ping, any news on this?

mikegrassotti:
I'm seeing the same problem - here's what I tried:
    git clone https://github.com/karmi/tire.git
    cd tire
    rvm use 1.9.2
    rvm gemset create tire-test
    rvm use @tire-test
    bundle install
    rake
    /Users/mgrassotti/.rvm/gems/ruby-1.9.2-p290@tired/gems/turn-0.8.3/lib/turn/autorun/minitest.rb:14:in `<top (require
d)>': MiniTest v1.6.0 is out of date. (RuntimeError)
`gem install minitest`and add`gem 'minitest' to you test helper.

karmi:
@grasscode, OK, thanks! So it's 1.9.2-related, since:

```
rbenv shell 1.9.3-rc1 
rake
LOADED SUITE test,test/integration,test/models,test/unit
...
```

Is there any pressing reason we should support 1.9.2?

coffeencoke:
Sorry for the lack of response, it's been a crazy couple of weeks.  @grasscode got the same thing I got.

coffeencoke:
@karmi no pressing reason, just something I think you should support someday.  We are using tire for our application, which is using ruby 1.9.2.

karmi:
It's of course stupidly easy to just add some line `gem "minitest" if RUBY_VERSION < '1.9.3'`... I was just wondering if 1.9.2 is widely used -- my impression is that people on 1.9.2. should migrate to 1.9.3?

karmi:
http://www.ruby-lang.org/en/downloads/ says:

>  The current stable version is 1.9.3.

I don't think using 1.9.2 in production makes sense, and do think supporting 1.8.7 and 1.9.3 is quite sufficient, so closing...

karmi:
So, any hard feelings if we close this as 1.9.3 seems better choice for development/production now?

kathgironpe:
+1 for 1.9.3.

coffeencoke:
Sounds good to me

Thanks!

Matt

karmi:
Great, thanks, closing then...

Issue #48, opened 2011-07-01
undefined method `paginate' for MyClass:Class, when importing


jpsilvashy:
So I'm using mongo, but I'm not sure thats the problem here, when I try to run `tire:import` on my model it fails with `undefined method`paginate' for Recipe:Class`

``` ruby
rake environment tire:import CLASS='Recipe' PARAMS='{:page => 1}' FORCE=1 --trace
```

``` ruby
** Execute environment
** Execute tire:import
# ...
--------------------------------------------------------------------------------
rake aborted!
undefined method `paginate' for Recipe:Class
```

And here is the stack trace for that:

``` ruby
rake aborted!
undefined method `paginate' for Recipe:Class
/Users/jpsilvashy/.rvm/gems/ruby-1.9.2-p180/gems/tire-0.1.14/lib/tire/index.rb:136:in `import'
Tasks: TOP => tire:import
```

Any ideas? I've tried it with and without `will_paginate` installed also with no success.

karmi:
OK, this was a tricky one. Yet, the solution is very simple:

``` bash
$ rake environment tire:import CLASS='Article.all' INDEX='mongo-articles' FORCE=true
```

Apparently, Mongoid does not support `Article.paginate`, but `Article.all.paginate`, for reasons beyond my understanding :)

Ping me if this works -- it does in my example app. _Tire_ does not care _what_ has the `paginate` method, only that _something_ in the app has it :)

karmi:
Ping, any news?

wulffeld:
With mongoid 2.0.2 will_paginate is no longer supported. That's what's causing this error in case that's not clear to anyone new coming here.

I've personally removed will_paginate and replaced it with kaminari which seems better. However, it has different semantics so it doesn't work with tire and stuff like:

``` ruby
Article.import :per_page => 100
```

breaks. I can't figure out if there's a better way around doing this:

``` ruby
Article.all.each do |conf|
  Article.elasticsearch_index.import [conf]
end
```

aaronchi:
This library needs a kaminari adapter... See
https://github.com/karmi/tire/issues/14

import just needs to use model.page if kaminari is installed for this to work correctly

karmi:
> I can't figure out if there's a better way (...)

Yes, there is:

``` ruby
Article.elasticsearch_index.import Article.all
```

as documented in the Readme.

> With mongoid 2.0.2 will_paginate is no longer supported. That's what's causing this error (...)

[...]

PR #151, opened 2011-11-16
Add ability to override the method used for finding objects when importi...

coffeencoke:
I would like to not have the dependency to paginate in order to use your import task.  We are using MongoDB which executes retrievals in batches natively.

### DIFF
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -9,6 +9,9 @@ namespace :tire do
           Pass params for the `paginate` method:
             $ rake environment tire:import CLASS='Article' PARAMS='{:page => 1}'
 
+          Use a method other than `paginate` (only use if your db operates batch retrievals natively):
+            $ rake environment tire:import CLASS='Article' FINDER_METHOD='all'
+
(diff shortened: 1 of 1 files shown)


coffeencoke:
sorry, this still does not work because the `import` method in lib/tire/index.rb forces `:page => options[:page]` which causes an exception for the `all` method.

What would you suggest we do?  When I run the import rake task we get the issue discussed in #48

lib/tire/tasks.rb
```diff
@@ -67,7 +71,7 @@ namespace :tire do
 
     STDOUT.puts '-'*tty_cols
     elapsed = Benchmark.realtime do
-      index.import(klass, 'paginate', params) do |documents|
+      index.import(klass, finder_method, params) do |documents|
```
karmi, line 74: I don't think this will work in your case. The `Index#import` method then relies on increasing the `params[:page]` variable. Does the batch finder in your case work the same way?

coffeencoke, line 74: Correct, I will take your advice from the other comment. 

Thanks!

Matt

PR #154, opened 2011-11-25
fixed rails new url

salax:
the old address redirected me to this url, rails new script couldn't handle that redirect


karmi:
Weird, works on Rails 3.1.1, merged & pushed though, thanks!

PR #156, opened 2011-11-29
Added min_score support to search

edmundsalvacion:
- Add min_score method to Search
- Added min score context to tests

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -65,6 +65,12 @@ module Tire
         self
       end
 
+      def min_score(value)
+        @min_score = value
+        @options[:min_score] = value
+        self
+      end
+
       def perform
         @response = Configuration.client.get(@url, self.to_json)
         if @response.failure?
@@ -93,6 +99,7 @@ module Tire
         request.update( { :size => @size } )               if @size
         request.update( { :from => @from } )               if @from
         request.update( { :fields => @fields } )           if @fields
+        request.update( { :min_score => @min_score } )     if @min_score
         request
       end
 
--- a/test/unit/search_test.rb
+++ b/test/unit/search_test.rb
@@ -344,6 +344,27 @@ module Tire
 
       end
 
+      context "min score" do
+
+         should "set the values in request" do
+          s = Search::Search.new('index') do
+            min_score 1
+          end
+          hash = MultiJson.decode( s.to_json )
+          assert_equal 1, hash['size']
+        end
+
+        should "set the min_score value in options" do
+          Results::Collection.any_instance.stubs(:total).returns(50)
+          s = Search::Search.new('index') do
+            min_score 1
+          end
+
+          assert_equal 1, s.options[:size]
+        end
+
+      end
+
     end
 
   end


karmi:
Hi, is it something you need or is it a "speculative" patch? :) Would you mind to create an extension to [_tire-contrib_](https://github.com/karmi/tire-contrib) for that?

edmundsalvacion:
Hi, it's something I need. I could certainly move it over to tire-contrib.

edmundsalvacion:
Question, do you have any guidelines on the modularity of items in tire-contrib?

Also, what do you see as appropriate to be in the main project vs contrib as far as the elasticsearch apis go?

karmi:
Well, if I understand the [ES docs](http://www.elasticsearch.org/guide/reference/api/search/min-score.html) correctly, it returns only a portion of results where the score is above a certain threshold. That _may_ just what you wanted in a tiny slice of use-cases. Most times, you want to actually display _some_ results...

The ES docs say: _Note, most times, this does not make much sense, but is provided for advance use cases_.

This seems to me as cross the line between _core_ and _contrib_: core is what 80% people would use, _contrib_ is for specific/advanced/crazy stuff.

karmi:
> Question, do you have any guidelines on the modularity of items in tire-contrib?

That's easy, just add/overload/etc anything in the Tire namespace, add tests, etc. See the [rails-logger](https://github.com/karmi/tire-contrib/blob/master/lib/tire/rails-logger.rb) extension for inspiration. The whole thing is modeled after [rack-contrib](https://github.com/rack/rack-contrib).

karmi:
Hello @edmundsalvacion, can I help you with the `tire-contrib` integration? Ping me here or at the #elasticsearch IRC!

It should be very easy to transfer the code from core to contrib. I'll close the pull request here and wait for a pull request on contrib...

timothyklim:
Hi, thanks for patch!

Why this patch is not included in core? Without this feature, search result is ugly(often).

+1 to core

karmi:
> Why this patch is not included in core?

Because it's a great fit to contrib.

> Without this feature, search result is ugly(often).

What do you mean?

timothyklim:
> What do you mean?

When you know that search result is not relevant when score below the value 0.5. Best way to hide irrelevant results is specify the "min_score".

> first day and first my usage experience of tire

After I've read tire manuals, I've started search "How to set minimum score?" and I found elastic guides and this pull request.

> That may just what you wanted in a tiny slice of use-cases. Most times, you want to actually display some results...

And we will have irrelevant results on the search page.

Thanks.

karmi:
> When you know that search result is not relevant when score below the value 0.5. Best way to hide irrelevant results is specify the "min_score".

I still consider the case where you don't want to present _any results whatsoever_ to the user an edge case. I can't imagine it widely used; the computed `score` varies wildly and people are usually better at judging the results, refining their query, etc.

I look forward to see the pull request on [`tire-contrib`](https://github.com/karmi/tire-contrib)!

sg552:
hi @Karmi, I am a Chinese user. Tire is great, but I didn't see the "min_score" function. Since in Chinese the street name is extremely complex to analyze. such as: "太阳宫", (which mean "sunny plaza") however without "min_score", most of the result is not relevant and below score 0.3.  so I think I need this option. 

I will commit the code to "tire-contrib". ( though I am monkeypatching the tire core ^_^ )

axsuul:
I don't see a pull request on tire-contrib for this. Should I do this?

Issue #162, opened 2011-12-07
Specify multiple nodes


coffeencoke:
Is there a way to specify multiple urls or hosts?

If our master node goes down, we want the ability to failover to another node.  If this does not exist we will probably be adding this functionality, so any direction from you would be great.

So far from browsing tire, I do not see anything that loads in a config yml file.  Is this correct?

karmi:
Hi, sorry for the delay.

**tl;dr: yes, it's probably a useful feature for `tire-contrib` and EC2 will eventually force us to implement it**

This is something which has been debated over and over, mainly on IRC with @clintongormley and others. It has even been said _Tire_ «misses the point of elasticsearch» because it does not handle high-availability in the client :)

Now, I have couple of things to say to this. Let me explain step by step.

First, let me say that I think that handling high-availability/failover in client “completely misses the point of elasticsearch” :) One of the _best_ things about elasticsearch is that it talks HTTP. Handling availability in HTTP stacks is what we, web developers, do all day.

Putting some kind of proxy (HAProxy, Varnish, custom Ruby proxy with [em-proxy](https://github.com/igrigorik/em-proxy)) in front of HTTP-based backend is easy, right? Most proxies can handle external sources for backend URL, can handle backend health, etc., right? That's what proxies are for, and people smarter then me have written those.

So, what you _should_ do, generally speaking, is to invest you intellectual energy into building a robust stack, where you put a proxy in front of the _elasticsearch_ cluster, which automatically does round-robin, health checks. Then you can talk to _elasticsearch_ from whatever client, even `curl`, and still have failover handling.

Obviously, it could be argued that you've just introduced a _single point of failure_ in your stack — but there's no free lunch.

Second, I've written one or two proxies in my life (see eg. https://github.com/igrigorik/em-proxy/blob/master/examples/balancing.rb), and while it's CS101-level programming, there are couple of interesting questions:
- How do you handle health checks? Do you introduce some background worker? How is the worker being run?
- How do you pass backend URLs to client? Do you introduce some kind of database?
- How do you handle dead backends from the client? Do you pass the info somewhere?

These are all questions you have to go over before you start writing code.

Third, let's review what's possible _now_, as we speak:

[...]

PR #163, opened 2011-12-08
Make it easier to re-perform a failed search on another server.

dylanahsmith:
Currently the url to make the search request to is set in the Search initializer.  This makes retrying a search without rebuilding the whole search not possible with the public Tire api.  This change allows the following to work:

```
search = Tire::Search::Search.new('articles')
# build search ...
servers = ["http://localhost:9200", "http://localhost:9201"]
begin
  Tire::Configuration.url servers[0]
  search.perform
rescue Errno::ECONNREFUSED => e
  Tire::Configuration.url servers[1]
  search.perform
end
# do something with results ...
```

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -4,16 +4,20 @@ module Tire
   
     class Search
 
-      attr_reader :indices, :url, :results, :response, :json, :query, :facets, :filters, :options
+      attr_reader :indices, :results, :response, :json, :query, :facets, :filters, :options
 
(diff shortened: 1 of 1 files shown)


karmi:
Hi Dylan, thanks! Added simple test and your example to commit message and pushed as karmi/tire@94e7b89.

Notice the issue karmi/tire#162, which would handle this failover scenario more conveniently -- I'll write some more info there once I'm done shaving fast breeding yaks with _Chef_'s `knife`.

PR #165, opened 2011-12-10
Docs & call for discussion on tire

ifesdjeen:
Hi Karel, hi Vojtěch!

Thanks a lot for an amazing work on Tire!

We're using ElasticSearch and Tire on several projects of ours. We switched over from Solr, and it was a deliberate decision. @michaelklishin from our team have recently added some integration specs for Mongoid. Now he's working more on a heavy lifting, so I had to take over Rails app and parts that involve search, too. 

Even though our our use-cases are not that complex, many things were not clear from Readme, so I went through the complete ElasticSearch doc and most of Tire source code and issues to understand some concepts (for example, nested models). Also, data types, and some other things. Since we used Solr, we were already familiar with analyzers and core concepts, so background knowledge helped. 

It's by no means a complete pull request, but while writing these things I discovered even more things I'd want to have in readme/wiki/doc, so I thought that I should talk to you guys before I proceed, since if you don't see that useful, I'll just write an internal doc for now.

My proposal would be to document and add even more integration specs for all three existing adapters and create a reference doc with some examples for people who were not yet familiar with full-text-search, give them some hints (like, that it's better to use "keyword" analyzer for things like tags and "snowball" for titles and article texts etc), maybe some more elaborate examples, and create something that will put together current readme, http://karmi.github.com/tire/ rocco doc and create a more complete wiki, with more examples and more information, which (of course) will partially duplicate one from elasticsearch.org. 

Let me know what you think. Since you guys created Tire, you know your final destination, but if you consider that helpful, I would be happy to keep pushing things to that pull request and get to know your thoughts!

Thanks!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -64,6 +64,12 @@ which allows you to use your preferred JSON library. We'll use the
     require 'yajl/json_gem'
 ```
 
+# Configuring objects (indexing)
+
+In order for ElasticSearch to find your objects, you have to specify which fields you want to store there,
+their type, analyzer and other things, which will improve quality of your search results. In order to
+inform tire about these things, you should use `Tire.index` method, or `mapping` class method on your models.
+
 Let's create an index named `articles` and store/index some documents:
 
 ```ruby
@@ -145,7 +151,52 @@ you can use _Tire's_ classes directly, in a more imperative manner:
     index.refresh
 ```
 
+## Field types
+
+For a complete list of types, complete and elaborate information upon each one of them, please refer
+[server documentation](http://www.elasticsearch.org/guide/reference/mapping/core-types.html).
+
+ElasticSearch provides following `string`, `date`, `boolean`, numeric (`integer`, `long`, `float` and
+`double`) and `binary`.
+
+## Analyzers
+
+For a complete list of Analyzers, complete and elaborate information upon each one of them, please refer
+[server documentation](http://www.elasticsearch.org/guide/reference/index-modules/analysis/index.html)
+
+In general analyzer in Lucene has toketizer, and token filter. Tokenizer splits text into chunks, filters
+process splitted text (for example, trunkate filter trunkates each token to specific length, stemmer extracts
+the word root, lowercase filter makes all characters in string lowercase etc.).
+
+Available analyzers are:
+  - [Standard](http://www.elasticsearch.org/guide/reference/index-modules/analysis/standard-analyzer.html),
+  - [Simple](http://www.elasticsearch.org/guide/reference/index-modules/analysis/simple-analyzer.html)
+  - [Stop](http://www.elasticsearch.org/guide/reference/index-modules/analysis/stop-analyzer.html)
+  - [Keyword](http://www.elasticsearch.org/guide/reference/index-modules/analysis/keyword-analyzer.html)
+  - [Pattern](http://www.elasticsearch.org/guide/reference/index-modules/analysis/pattern-analyzer.html)
+  - [Language](http://www.elasticsearch.org/guide/reference/index-modules/analysis/lang-analyzer.html)
+  - [Snowball](http://www.elasticsearch.org/guide/reference/index-modules/analysis/snowball-analyzer.html)
+
+You can also create a [custom](http://www.elasticsearch.org/guide/reference/index-modules/analysis/custom-analyzer.html)
+analyzer by combining any of he available Tokenizers and Filters.
+
+Even more information on analyzers is available in [Lucene project](http://lucene.apache.org/java/3_4_0/api/all/index.html),
+which powers ElasticSearch, too.
+
+## Boost
+
+Boosting helps you to control relevance of documents. For example, you can specify that matching of a :title
+field is more important than matching :body field in case with articles. Terms of the full-text search
+query can also be boosted, so you can specify which terms are more important, so that they could get a better
+(or worse, in case of the negative boost), ranking.
+
+By default boost factor is 1. It should always be positive, but may be less that 1 (for example 0.5). The
+higher the boost factor is, the more relevant the field / term will be.
+
+
 OK. Now, let's go search all the data.
+
+## Search
 
 We will be searching for articles whose `title` begins with letter “T”, sorted by `title` in `descending` order,
 filtering them for ones tagged “ruby”, and also retrieving some [_facets_](http://www.elasticsearch.org/guide/reference/api/search/facets/)
--- a/README.markdown
+++ b/README.markdown
@@ -154,10 +154,40 @@ you can use _Tire's_ classes directly, in a more imperative manner:
 ## Field types
 
 For a complete list of types, complete and elaborate information upon each one of them, please refer
-[server documentation](http://www.elasticsearch.org/guide/reference/mapping/core-types.html).
+[ElasticSearch guide](http://www.elasticsearch.org/guide/reference/mapping/core-types.html).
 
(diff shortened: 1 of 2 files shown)


karmi:
Hi, thanks!

I've got two things to say first:

1/ The main focus of _Tire_ is to make it incredibly easy for Rubyists to start with elasticsearch. It may be evident from the way the Readme is written, from the fact that you can generate a fully working Rails application with one command, and other things.

2/ The second focus of _Tire_ is to make it convenient for Rubyists to make use of advanced _elasticsearch_ features. It may be evident when you consider how boolean queries take blocks, how _Tire_ does not trample on your models, and other things.

Now, consider how you write in your patch:

> In order for ElasticSearch to find your objects, you have to specify which fields you want to store there,
> their type, analyzer and other things (...)

That's simply not true. You **don't have to specify** anything in _elasticsearch_. It works out of the box, and its author went to great lengths to make it happen.

Notice also, how you introduce the `mapping` method before we even started talking about putting some simple data into _elasticsearch_. I think the Readme has some flow to it, and continually builds more advanced stuff from simple stuff.

Now, field types, analyzers, boosts? Before we even started talking about search? Why? There's no reason to do that in _elasticsearch_ — it **just works**. You download & untar some file from teh internets, run shell script, and can work with it.

Notice also, how you start including pointers to _elasticsearch_ documentation early on. Again, why? Does a buddy Rubyist wanting to try out this new hipster search engine really have to open twenty tabs in her browser to start working with _Tire_? I don't think so.

The problem with documentation, generally, is that people either complain it's too thin or too thick. I think both _Tire_ documents have some balance, and if a new user gives them an hour or so, she can start doing pretty powerful stuff with _elasticsearch_ right away.

Obviously, after the initial crush is over, you **will have to look at _elasticsearch_ documentation**, no matter what. You'll discover that there are many more different APIs, many analyzers, that you actually don't understand anything, etc. But that's good — by that point you know if you like _elasticsearch_ or not. My bet is you do.

What to do about _Tire_'s documentation, then? I'm all ears for suggestions how to make both main documents better, but, generally, they work for me.

You are very right when you say:

> create a reference doc with some examples for people who were not yet familiar with full-text-search,
> give them some hints (like, that it's better to use "keyword" analyzer for things like tags and "snowball"
> for titles and article texts etc), maybe some more elaborate examples (...)

We've been already debating it with @michaelklishin couple of weeks ago, and hell yes, I'm enthusiastic about this. A wiki page which says, “OK, guys, here's the story how you should index articles with tags” and a page which says “here's the story how you should index and setup everything for an address book”.

That would be awesome, because then you could demonstrate stuff like “use `keywords` analyzer for tags” and “make sure you use `multifield` for authors, so you can fulltext search them and also do facets on them”, or stuff like “here's how you should store phone numbers” and “here's how you should store people's resumes in PDF in the index”.

We've already agreed with Michael that the best way is to just create a Wiki page(s) for stuff like this, and work on the _content_. Once we have the content, it's trivial to make decisions how it should be structured, how it should be splitted, etc. Whether it stays on the Wiki or whether we build some more extensive website is another decision — but we have to have excellent content _first_.

And of course, this is also about @elasticsearch's documentation itself — we are constantly revising that, and everybody knows it should get better, both in content and structure. I'm pretty confident sooner then later it will get substantially better.

Some of the stuff you have written really “partially duplicates” _elasticsearch's_ own documentation, and that's not good. We should contribute upstream, to the main project, not scribble on the edges of _elasticsearch's_ documentation.

ifesdjeen:
I'm not sure wether duplicating data is generally good or bad. It's definitely harder to support, but taking a wise approach, maintenance is minimized or eliminated. I certainly think that contributing to upstream is #1 goal, although sometimes giving a quick hint and a couple links to the upstream still seems good to me.

So far only thing I could tell is that reading though the readme with quite general headers (Installation / Usage / Integration), basically covering everything. Rocco docs were always difficult for me to read, because of line breaks and jumps, but it's just me. 

My copy is certainly not best one (i often find myself writing "have to" instead of "have an option"), and order of things may certainly vary. Moreover, these things like boost, attributes and links to analyzers, tokenizers and other things may be not somewhere on the top, but not mentioning them is not the best thing to do as well (imo). Even though ES works out of the box, knowing how to configure things that most everyone requires in his app sounds good. 

From the rails app developer standpoint, when you develop a prototype of an app, you don't really have any data in app. So you often think of how data is indexed before you get the data. While your test data is added, hooks are called and it's indexed. I don't really know what's the percentage of the existing vs new apps that start using Tire though, but in my practice, I was always first creating mappings and indexes, and only later - cleaning up indices and re-creating them with new options or mappings. 

Of course, my thinking was not to get whole elasticsearch.org to the readme. Rather - showing connections between Tire and ES (if you specify data type, go take a look at which ones are there, if you put an analyzer at least get to know what they mean, what attributes mean and where to look them up, how to specify multiple sort field and create more complex queries etc).

I'm just a user, and I was completely fine with going to elasticsearch.org and checking out some more complex things. Although I thought that it may be good to give a couple of iterations to readme / wiki to make some things more obvious without checking out source code and test suite. 

I'll keep thinking of the way to present that info. I'm starting work on Clojure driver, little by little. That may help me to understand things in a better way.

Thanks for explanation.

karmi:
Hey Alex,

> Even though ES works out of the box, knowing how to configure things that most everyone requires in his app sounds good. (...)

yes, but I can't see how we can do all that in the Readme — that's why I tried to give you feedback.

I think the most important point here is: if you and Michael would be willing to write the Wiki pages, as we have debated, **that** would help many users, absolutely. Nothings beats a well written, well prepared example...

ifesdjeen:
Agreed. 
We'll take a couple of examples and work on organizing them in an explanatory way. 
Whenever we have something to show, we'll ping you for a review. 

Thanks a lot for helping out!

karmi:
Perfect, just put them in the Wiki, that would be awesome.

michaelklishin:
I have never seen a project that puts half of documentation it has in the README and is known for its docs. You cannot please everyone with one document. I haven't really read contributed changes in full but just taking them as one document in the wiki will produce the same result: either too much information all at once or only providing some very basic examples that will lead people to believe that Tire and ES cannot handle advanced cases.

karmi:
Michael, I never suggested you create **one Wiki page** for the examples which we have discussed. Obviously, each example could get its own Wiki page. Once we have the content, we can debate how to structure and present it better.

(Not to start any clashes :), but, the only documentation for @sinatra was initially, and for a long time, in the Readme. The general feedback seemed OK. I always loved it. Around the 1.0 release @rtomayko created a website and wrote the awesome content such as http://www.sinatrarb.com/extensions.html. The current form of the Readme is much less awesome then I remember from the old days -- but still, it's on one page.)

michaelklishin:
I agree that it worked well for Sinatra. Sinatra is different in two ways:
- People who came around it were already familiar with web development in some shape or form. Not the case for search servers like ES or Solr, especially given the widespread "Java is hard and ugly and always means overengineered crap" thinking in the Ruby community.
- Sinatra has (intentionally) next to no advanced features. ES has plenty and thus Tire will have support for at least some of them (ideally, all).

But I see what you are saying and my intent is to begin adding wiki pages in the same order [amqp gem guides](http://rubyamqp.info) are structured, then the README can be trimmed down.

karmi:
Yes, agreed. And yes, something like http://rubyamqp.info/articles/patterns_and_use_cases is precisely what I think would benefit _Tire_ tremendously.

As for the Readme, sure, I like the structure it has, but obviously it contains too much info -- in the ideal world, it would have links “More about mapping”, “More about ActiveRecord”, à la http://gembundler.com.

PR #171, opened 2011-12-13
Reraise RequestTimeout and ServerBrokeConnection exceptions. Fixes #170

### LINKED ISSUE #170: RequestTimeout and ServerBrokeConnection are caught
Tire is catching these exceptions in its http client, and returning a response with an error code of 0 and http body of nil.  However, these are indicators of a connection errors.

To reproduce these exceptions, start out by running `nc -l 9200` in one terminal, then use Tire.configuration.client to make a request in another terminal, then
1. for RestClient::RequestTimeout just wait for the request to timeout.
2. for RestClient::ServerBrokeConnection just kill nc (Ctrl-C, Ctrl-D or Ctrl-).

The expected result is that the exception is raised, so that it can be handled along with other connection errors for failover, as discussed in issue #162.

dylanahsmith:


karmi:
Dylan, thanks, very nice patch -- could you please add a small test to https://github.com/karmi/tire/blob/master/test/unit/http_client_test.rb that it in fact does re-raise the exception?

dylanahsmith:
@karmi test added.

karmi:
Thanks Dylan, merged, pushed. I had some strange `StackLevelTooDeep` errors, so I simplified the test suite and kinda cleaned it up.

Actually, a question: have you tried to use Curb at Shopify? It should be noticeably faster under concurrent load.

PR #175, opened 2011-12-17
Support setting multiple URLS in the configuration

cover:
Currently Tire supports just one ElasticSearch url, it would be useful to support more and let it randomly use one of the available urls on requests (some kind of failover would be nice too, maybe next)

### DIFF
--- a/lib/tire/configuration.rb
+++ b/lib/tire/configuration.rb
@@ -7,6 +7,7 @@ module Tire
     end
 
     def self.url(*values)
+      values.flatten!
       @urls = values.map{|value| value.to_s.gsub(%r|/*$|, '')} if values.any?
       urls.respond_to?(:sample) ? urls.sample : urls.choice
     end
--- a/lib/tire/configuration.rb
+++ b/lib/tire/configuration.rb
@@ -2,13 +2,13 @@ module Tire
 
   class Configuration
 
-    def self.urls(values=nil)
-      @urls = (values && values.any? ? values.map{|value| value.to_s.gsub(%r|/*$|, '')} : nil) || @urls || ["http://localhost:9200"]
+    def self.urls
+      @urls || ["http://localhost:9200"]
     end
 
-    def self.url(value=nil)
-      urls([value])
-      @urls.respond_to?(:sample) ? @urls.sample : @urls.choice
+    def self.url(*values)
+      @urls = values.map{|value| value.to_s.gsub(%r|/*$|, '')} if values.any?
+      urls.respond_to?(:sample) ? urls.sample : urls.choice
     end
 
     def self.client(klass=nil)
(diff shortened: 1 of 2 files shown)


karmi:
Hi Fabio, thanks for the patch, I'll have a look at it in the light of proposed changes at karmi/tire#162.

cover:
Hi Karmi, damn, I didn't see that one, I'm reading it right now :)

cover:
After reading it I've changed something, now it's possible to do
Tire.configuration { url 'http://es1.example.com', 'http://es2.example.com' }

I kept the sample/choice in order to return one random url from the ones available and not breaking anything, at least until there will be the round-robin support.

cover:
...and...

Tire.configuration { url ['http://es1.example.com', 'http://es2.example.com'] }

which could be more realistic with something like:

elasticsearch = YAML::load( File.open( Rails.root.join('config', 'elasticsearch.yml') ) )
Tire.configuration { url elasticsearch[Rails.env][:hosts] }

karmi:
@karmi @vhyza Bump!

speedmax:
hopefully this gets merge into master soon

nvdk:
just wanted to add my interest for this option here :)

krae:
Yah, same. Are you still looking at this?

lib/tire/configuration.rb
```diff
@@ -2,8 +2,14 @@ module Tire
...
-    def self.url(value=nil)
-      @url    = (value ? value.to_s.gsub(%r|/*$|, '') : nil) || @url || "http://localhost:9200"
+    def self.urls
+      @urls || ["http://localhost:9200"]
+    end
+
+    def self.url(*values)
+      values.flatten!
+      @urls = values.map{|value| value.to_s.gsub(%r|/*$|, '')} if values.any?
+      urls.respond_to?(:sample) ? urls.sample : urls.choice
```
karmi, line 12: Hey, I'm a bit uncertain how this would work with the HTTP clients. The methods accepts an array, sets the `@urls`, and the returns a random item from the array, correct?

I think we need much more robust implementation, as outlined in #162.

What I like about the implementation is that everybody can simply stick this:

``` ruby
module Tire
  class Configuration

    def self.url(*values)
      values.flatten!
      @urls = values.map{|value| value.to_s.gsub(%r|/*$|, '')} if values.any?
      urls.respond_to?(:sample) ? urls.sample : urls.choice
    end

  end
end
```

into their `lib` and they get the behaviour. I'm leaning towards closing this and focusing on the more robust implementation with round robin pooling, graceful handling of failed IPs, auto-discovery of cluster nodes a la the Perl client, etc. What do you think?

karmi:
Closing in favour of a more complete implementation.

PR #185, opened 2011-12-23
Lazy loading of Searches only when #results or #response called

gedwards:
I had a use case in the DSL where a previously created query needed to be modified.

search = Tire.search 'test' do query { string 'some_field:some_value' } end   # don't perform the query yet

search.filter :terms, invite: ['yes']    # add an additional filter

search.results   #now execute the query (so that query is executed only once)

This commit adds this ability (it was easy to do) and includes specs. As a bonus, it replaces many instances of ".perform.results" with the more simple ".results"

### DIFF
--- a/lib/tire/dsl.rb
+++ b/lib/tire/dsl.rb
@@ -7,7 +7,7 @@ module Tire
 
     def search(indices=nil, options={}, &block)
       if block_given?
-        Search::Search.new(indices, options, &block).perform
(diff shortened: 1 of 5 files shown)


karmi:
Greg, thanks! An _awesome_ commit and a great improvement! Merged in as karmi/tire@9196141

(Also see karmi/tire@0b65457)

Issue #184, opened 2011-12-23
Better DSL for custom attributes


ream88:
At the moment Tire isn't very flexibly when custom attributes are needed for searching. I'm often ending up mixing `mapping` and `to_indexed_json`. One model doesn't use custom attributes so it defines its searchable attributes via `mapping` – which is imho the better and far more beautiful way –, another models need one stupid custom attribute and I'm forced to use `to_indexed_json`. It's not really a problem, but it just doesn't feel right.

How about implementing something like this:

``` ruby
class Article
  belongs_to :author

  mapping do
    indexes :author, as: proc { author.name }
    indexes :author_email, as: 'author.email'
  end
end
```

My idea is to implement another option for the `indexes` method, maybe `:as` which accepts a lambda/proc or a string which will be evaluated at indexing time. Any thoughts?

PR #186, opened 2011-12-23
Implemented `:as` mapping option

ream88:
Implemented `:as` mapping option, see #184 for more details.


karmi:
Hi, thanks!, was thinking about adding something like that and you did it nicely! Merged, added some tests & docs, pushed!

ream88:
Thank you for the quick pull and the really really great gem!

PR #189, opened 2011-12-28
A couple of fixes for eager loading

ddnexus:
Sorry, there was a bug in my previous pull-requested-commits, so here is the right one.

Tire.search (%w[red_things blue_things], :load=>true) do ...

Returns a mixed type collection that cannot be loaded by the current implementation. The first patch fixes it.

Besides, if you use nested model names (e.g. 'This::Thing') the model_name.singular returns a default document_type that is not usable to reconstruct the original class. Using 'underscore' instead of 'singular' fixes it.

### DIFF
--- a/lib/tire/model/naming.rb
+++ b/lib/tire/model/naming.rb
@@ -74,7 +74,7 @@ module Tire
         #
         def document_type name=nil
           @document_type = name if name
-          @document_type || klass.model_name.singular
+          @document_type || klass.model_name.underscore
         end
       end
 
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -39,23 +39,32 @@ module Tire
             end
           else
             return [] if hits.empty?
-
-            type  = @response['hits']['hits'].first['_type']
-            raise NoMethodError, "You have tried to eager load the model instances, " +
-                                 "but Tire cannot find the model class because " +
-                                 "document has no _type property." unless type
-
-            begin
-              klass = type.camelize.constantize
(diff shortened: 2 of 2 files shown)


karmi:
Hi, thanks for having a look into this... Alas, there's already a pull request (and quite a long discussion) about multi-model searches here: https://github.com/karmi/tire/pull/131 -- could you have a look, please? (I still think the `results` method must be first refactored, to allow more acrobatics like these...)

I'll definitely have a look into the namespacing issue!

ddnexus:
My patch is actually dead simple and doesn't add any burden to the method. It just keep into consideration that the models could be different, and just load them all with one find call per model.

About the comments in the posts you are linking: IMHO it's a matter of (in)consistency. :-)

If you support the :load option, you should fix it and make it work consistently with the rest of your code, i.e. multiple types objects. On the other hand If you consider that loading is "suspicious", you could remove the option completely, so at least it will not generate an error. Keeping the option and not supporting it for that special case makes me dizzy :-), specially because the fix is so simple.

The other patch for the default document_type fixes the same type of inconsistency: the (Tire) code of one method is inconsistent with the usage that the same (Tire) code does somewhere else. 

klass.model_name.singular is not suitable to generate any document_type that will work with camelize or classify, while uderscore does just that.

karmi:
Hi, thanks for the naming patch -- I've merged, added a test for the behaviour and pushed.

(As for the other thing, feel free to submit another ticket, if you like. I don't have anything new to add to the discussion at the moment.)

PR #190, opened 2011-12-29
Search with version should set the version value in options

gedwards:
Adding 'version' to the DSL. Includes unit and integration tests. (thanks for Tire and Elasticsearch!)

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -69,6 +69,10 @@ module Tire
         self
       end
 
+      def version(value)
+        @version = value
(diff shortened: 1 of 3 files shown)


karmi:
Greg, thanks, the delay with merging is embarrasing, sorry. Pushed.

(By the way, I think a better way would be an API like `Tire.search 'an_index', version: true do ... end`, but it does not matter at the moment. We have to sort out the issue with search params anyway, and transfer the version option there once it's done...)

PR #191, opened 2011-12-30
Kaminari-compatible import rake task

chromeragnarok:
Hi guys,
I've written a patch so the import rake task will support Kaminari out of the box, without adding the 'paginate' implementation into the model objects. Thanks a lot!


karmi:
Yes, why not do this, actually :) Merged & pushed, thanks!

chromeragnarok:
You're welcome :D

PR #193, opened 2012-01-06
Return nil from Index#retrieve when the document doesn't exist.

dylanahsmith:
Previously an exception was raised on document.update(h['fields']) since h['field'] returned nil.

Here is a simple shell session that shows the data returned when a document is missing:

``` shell
$ curl -XPOST localhost:9200/empty
{"ok":true,"acknowledged":true}
$ curl -XGET localhost:9200/empty/message/1
{"_index":"empty","_type":"message","_id":"1","exists":false}
```

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -152,10 +152,9 @@ module Tire
       h = MultiJson.decode(@response.body)
       if Configuration.wrapper == Hash then h
       else
-        document = {}
-        document = h['_source'] ? document.update( h['_source'] ) : document.update( h['fields'] )
+        document = h['_source'] || h['fields'] || {}
         document.update('id' => h['_id'], '_type' => h['_type'], '_index' => h['_index'], '_version' => h['_version'])
-        Configuration.wrapper.new(document)
+        h['exists'] != false ? Configuration.wrapper.new(document) : nil
       end
 
     ensure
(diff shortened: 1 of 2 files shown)


lib/tire/index.rb
```diff
@@ -152,10 +152,9 @@ def retrieve(type, id)
       h = MultiJson.decode(@response.body)
       if Configuration.wrapper == Hash then h
       else
-        document = {}
-        document = h['_source'] ? document.update( h['_source'] ) : document.update( h['fields'] )
```
karmi, line 156: Why not simply do a?

```
document = h['_source'] && h[exists] ? ... : ...
```

(Sorry, never encountered this, so I'm trying to understand the issue.)

dylanahsmith, line 156: The issue is, then when I try to retrieve a document that doesn't exist, it throws an error.

Try running the following:

``` ruby
require 'rubygems'
require 'tire'
index = Tire::Index.new("empty")
index.create
puts index.retrieve("doc", 1).inspect
```

And you will get an error as follows:

```
/Users/dylansmith/src/tire/lib/tire/index.rb:156:in `update': can't convert nil into Hash (TypeError)
    from /Users/dylansmith/src/tire/lib/tire/index.rb:156:in `retrieve'
    from test.rb:7:in `<main>'
```

You're suggestion doesn't help, because it will still result in `document.update(nil)` being executed.

karmi, line 156: Yeah, true.

karmi:
Thanks, merged & pushed!

PR #194, opened 2012-01-06
Implement put mapping API.

dylanahsmith:
http://www.elasticsearch.org/guide/reference/api/admin-indices-put-mapping.html

The mapping for all types can be set in the index creation request, but
the mapping for a single type must be updated per put mapping request.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -41,6 +41,20 @@ module Tire
       MultiJson.decode(@response.body)[@name]
     end
 
+    def update_mapping(mapping)
+      mapping.each do |type, value|
+        url, body = "#{Configuration.url}/#{@name}/#{type}/_mapping", MultiJson.encode(type => value)
+        begin
+          @response = Configuration.client.put url, body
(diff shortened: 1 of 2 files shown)


karmi:
Hello Dylan, thanks, sorry for such embarrasing delay.

I hate to be the asshole who asks for more tests all the time, but I think an integration test would be warranted?

Also what about mapping conflicts? What about the `ignore_conflicts` option?

jfredett:
Pretty sure this would fix index creation w/ bonsai (with a little bit of supporting code) -- I'll see if I can write up some integration tests when I get it all working w/ bonsai. 

Open source at work!

nz:
+1; this is relevant to my interests. Updating mapping at the index level won't be an issue with Bonsai for much longer, but it always helps to be specific.

@jfredett feel free to drop me a line if you need any support for this stuff on our end over at Bonsai

karmi:
@dylanahsmith Let's close this until we have a more complete iplementation with handling or ignoring conflicts, displaying diffs, etc etc?

PR #198, opened 2012-01-12
Add minitest gem and search_type support

erickt:


karmi:
@erickt What do you think about the API? I think at least three people are now trying to add various request params to Tire, without much communication. It would be awesome if we could sort it out here in issues or at the IRC channel...

erickt:
Hi Karmi! Sorry for the absence, I've been busy with non-tire related things. I poked around a little more, and it looks like the `minitest` dependency is coming from the gem `turn`, but the `turn` gem only has a development dependency on `minitest`, which bundler seems to ignore (at least for me).

erickt:
Regarding the api, I'm divided on what to support. First off, do you actually want the non `search_type` uri parameters to get passed by a uri parameter. `timeout`, `from`, and `size` can all be passed in the json body. Second, would you only want the parameters to be passed in the `Search#new` function, or do you also want the dsl functions? I could see it being a little confusing if you write:

```
s = Search::Search.new('index', size: 10) do
  size 20
  query { string 'title:foo' }
end
```

Which `size` declaration wins?

PR #204, opened 2012-01-14
Default document_type and index_name implementation 

ddnexus:
This simple convention/conversion should be AR, ruby and URL friendly. It also keeps the index_prefix into consideration in case you want to classify an index with a prefix.

### DIFF
--- a/lib/tire.rb
+++ b/lib/tire.rb
@@ -5,6 +5,7 @@ require 'hashr'
 
 require 'tire/rubyext/hash'
 require 'tire/rubyext/symbol'
+require 'tire/rubyext/string'
 require 'tire/logger'
 require 'tire/configuration'
 require 'tire/http/response'
--- a/lib/tire/model/naming.rb
+++ b/lib/tire/model/naming.rb
@@ -30,7 +30,7 @@ module Tire
         def index_name name=nil, &block
           @index_name = name if name
           @index_name = block if block_given?
-          @index_name || [index_prefix, klass.model_name.plural].compact.join('_')
+          @index_name || [index_prefix, klass.model_name.tableize.gsub('/', '__')].compact.join('_')
         end
 
         # Set or get index prefix for all models or for a specific model.
@@ -74,7 +74,7 @@ module Tire
         #
         def document_type name=nil
           @document_type = name if name
-          @document_type || klass.model_name.singular
+          @document_type || klass.model_name.underscore.gsub('/', '__')
         end
       end
 
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -46,7 +46,7 @@ module Tire
                                  "document has no _type property." unless type
 
             begin
-              klass = type.camelize.constantize
+              klass = type.document_type_classify
             rescue NameError => e
               raise NameError, "You have tried to eager load the model instances, but " +
                                "Tire cannot find the model class '#{type.camelize}' " +
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -46,7 +46,7 @@ module Tire
                                  "document has no _type property." unless type
 
             begin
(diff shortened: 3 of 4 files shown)


karmi:
Hi Domizio, thanks for keeping looking into this. There are two problems, however, with the patch:
1. It has no tests attached, which makes it very hard to include -- I'd have to write them myself.
2. Given there are no tests, I'm not entirely sure I understand how it works -- you change the `/` to `__` when you pass data in, but where do you do the inverse when getting them out?

Also, I think all of this is doable without extending the `String` class...

ddnexus:
It was just a possible suggestion... for that reason I didn't attach any test... however, if you like the patch I could. Should I?

The goal was to show that a simple convention is enough to work around the restriction imposed by the usage of index_name and document_type in the url. Indeed we don't need a class with #to_s and #to_url methods: we can just avoid using the slash, by gsubbing it to some other value, like a double '__'.

About what the patch does, I wrote you in the original starting post for the first patch... I paste here something from the other post:

```
>> MyNested::SpecialModel.index_name
#=> "my_nested__special_models"

>> MyNested::SpecialModel.document_type
#=> "my_nested__special_model"

>> "my_nested__special_model".document_type_classify
#=> "MyNested::SpecialModel"

>> "my_nested__special_models".index_classify
#=> "MyNested::SpecialModel"
```

besides, you can also do:

```
class MyNested::WeirdModel
    index_prefix 'my_prefix'
end

>> MyNested::WeirdModel.index_name
#=> "my_prefix_my_nested__weird_models"

>> "my_prefix_my_nested__weird_models".index_name_classify
#=> "MyNested::WeirdModel"
```

About extending String: I agree it's not strictly necessary, but it is just in style with all the other String methods extended by ActiveSupport so we could just follow the same convention, along with the other String added methods:
- pluralize
- singularize
- constantize
- camelize
- titleize
- underscore
- dasherize
- demodulize
- parametrize
- tableize
- classify
- humanize

In our case the additions to string are very Tire specific due to their name that is composed by the original method name, and the '_classify' suffix, and we don't override anything that is already in place, so I would use that instead a not-so-elegant-and-against-rails-convention function.

I should check where (and whether) Tire converts document_type (and/or index_name) back to the class name. One place that I was aware of is the `tire/results/collection.rb` so I patched it there. Anyway, although not used by Tire directly or extensively, the methods are useful to users, specially if you are going to drop the :load option or encouraging users to write their own code around Tire.

If you provide a default from class name to index_name and document_type, I would expect to have a canonical method that allows me to convert them back to the right class name, and possibly doing it with the standard way ActiveSupport made me used to. (i.e. string method)

rubish:
+1

Just got bit by this.

karmi:
Hi,

yes, without tests, it's really hard to see what's the intended purpose and cases.

I wouldn't extend `String`, and would like some symmetry in the code, so the encoding `document_type_classify` has its decoding counterpart and it's not added into `index_name` method.

ddnexus:
Hi Karel,

> I wouldn't extend String,

great!  :-)

> the encoding document_type_classify has its decoding counterpart

I cannot understand the meaning of this one. Maybe it's the "encoding/decoding" concept that is not clear to me. If it can be interpreted as "translate a string from one convention to the other"... do you mean something like `document_type_to_class_name(doc_type_str)` to return the class name given the `document_type`, and `class_name_to_document_type(class_name)` to return the document type given the `class_name`?

> and it's not added into index_name method.

Again, I am confused with this one. The 'into' part is what make me think that I probably don't understand what you mean. Do you mean that the `index_name` should not have the same kind of translation/encoding, or that you don't want to use it 'inside' the index name method? 

Providing a default algorithm to generate the `index_name` (that is also including a prefix) should be coupled with a "decoding counterpart" (?) so when you are dealing with an index_name you can get the class name using the opposite algorithm. That is quite useful in order to eliminate the need of mapping hashes, for example.

PR #210, opened 2012-01-19
Support any level of mappings

woahdae:
Previously, only two levels of mappings were supported, ex:

```
mapping do
  indexes :title, :type => 'string'
  indexes :author do
    indexes :first_name, :type => 'string'
    indexes :last_name,  :type => 'string', :boost => 100
  end
end
```

This commit adds support for any level of mappings, ex:

```
mapping do
  indexes :title, :type => 'string'
  indexes :author do
    indexes :first_name, :type => 'string'
    indexes :last_name,  :type => 'string', :boost => 100

    indexes :posts do
      indexes :title, :type => 'string', :boost => 10
    end
  end
end
```

### DIFF
--- a/lib/tire/model/indexing.rb
+++ b/lib/tire/model/indexing.rb
@@ -76,21 +76,21 @@ module Tire
         # for more information.
         #
         def indexes(name, options = {}, &block)
+          mapping[name] = options
+
           if block_given?
-            mapping[name] ||= { :type => 'object', :properties => {} }.update(options)
-            @_nested_mapping = name
-            nested = yield
-            @_nested_mapping = nil
-            self
-          else
-            options[:type] ||= 'string'
-            if @_nested_mapping
-              mapping[@_nested_mapping][:properties][name] = options
-            else
-              mapping[name] = options
-            end
-            self
+            mapping[name][:type]       ||= 'object'
+            mapping[name][:properties] ||= {}
+
+            previous = @mapping
+            @mapping = mapping[name][:properties]
+            yield
+            @mapping = previous
           end
+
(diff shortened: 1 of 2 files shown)


digitalplaywright:
@woahdae Great work! This is a very useful feature. However, when I tested this it does not seem to support the ':as' option like e.g 
   mapping do
     indexes :all_attendees, :as => :attendees
   end

Every field that is indexed with the :as option is not persisted to the elastic search index.

Are you able to reproduce the problem with ':as'?

woahdae:
The full test suite passes, and I'd hope :as has coverage, but I'll give it a check in my app.

digitalplaywright:
The problem I reported seem to be present on master as well: 

':as => :function_name' does not work, but ':as => 'function_name' works.

digitalplaywright:
@karmi This works for us in production, and it is essential for putting our Mongoid documents into an ElasticSearch index. Please let me know if there are any plans to merge this with master and if not what stands in the way of this.

Thank you.

karmi:
@digitalplaywright Yeah, I'll have a look into it tommorrow.

@woahdae Sincere apologies!, I like the patch very much, sorry it got stalled like this...

karmi:
@digitalplaywright The issue you're having is due to https://github.com/karmi/tire/blob/master/lib/tire/model/search.rb#L169. That's easy to fix :)

karmi:
@woahdae Hi, thanks _very much_ for patch, finally merged! Sorry it took so long...

digitalplaywright:
@karmi Thank you for merging this and for all your hard work on Tire.

PR #211, opened 2012-01-20
Add support for Boolean Filtered Queries

leehambley:
- http://www.elasticsearch.org/guide/reference/query-dsl/bool-filter.html

This patch only implements the `must` logic, and has no facility for the
caching behaviour offered by boolean filtered queries.

This is slightly different to the `filter(:and, { {filter1},
{filter2}})` option introduced in `d2feb840` by @karmi in that it has
different semantics, and doesn't require one to collect a list of
filters (for example in conditional blocks) before calling
`filter(:and,...)`

This patch is incomplete, I will implement `should` and `should_not` as
soon as time becomes available. I don't expect it to be pulled upstream
until those are implemented, but perhaps there is enough value in my
contribution that it will be pulled.

### DIFF
--- a/lib/tire/search/facet.rb
+++ b/lib/tire/search/facet.rb
@@ -17,7 +17,7 @@ module Tire
       def terms(field, options={})
         size      = options.delete(:size) || 10
         all_terms = options.delete(:all_terms) || false
-        @value = if field.is_a?(Enumerable)
+        @value = if field.is_a?(Enumerable) and not field.is_a?(String)
           { :terms => { :fields => field }.update({ :size => size, :all_terms => all_terms }).update(options) }
         else
           { :terms => { :field => field  }.update({ :size => size, :all_terms => all_terms }).update(options) }


leehambley:
@karmi any thoughts on this, first-pass style-wise? Should I finish my work (trivial) and add `should` and `should_not`

karmi:
@leehambley I'm terribly sorry for this embarrasing delay :(

I think it's a valid approach and the code is very fine. Nevertheless, could you please get in touch with @ralph, and check with him how this plays with the recently merged support for filtered queries? See https://github.com/karmi/tire/blob/master/test/integration/filtered_queries_test.rb#L43-60 for the API.

If I'm not completely mistaken, it does allow the same sort of thing, with a little bit different API. I tend to overzealously guard _Tire_ against more lines of code being introduced -- I've said numerous times that the codebase feels too heavy to me already and would like to make it slimmer... 

Do not read this as opposition to the patch and the proposed API -- I just think that if we can achieve the same thing with already existing features, we should maybe be a bit asketic here...

ralph:
Hey, I will go on a 10 days vacation today, but I will get back to Lee later. As we are in one city, we'll probably just meet and hack on this together.

leehambley:
Whoohoo, viel Spaß beim Urlaub! Looking forward to hacking when you are
back, much easier than GH issues! @karmi, we'll come back to you!

On 2 March 2012 09:13, Ralph von der Heyden <
reply@reply.github.com

karmi:
@ralph , @leehambley That sounds like a great plan, thanks! :) Enjoy the vacation, Ralph!

PR #218, opened 2012-01-30
ActiveRecord Single Table Inheritance search bug

sandrew:
When I need to make a search on STI ancestor (to get mixed results from all descendants) on one index (like in tests), I was usually getting error, because it makes 'find' from wrong class. Well, you can see it in tests.

Also it repaires one more bug: when you have two STI classes with separate indexes (like in default) and try to make a search like Tire.search('first_table_index,second_table_index', :load => true) { ... }, it also raises RecordNotFound.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -40,24 +40,25 @@ module Tire
           else
             return [] if hits.empty?
 
-            type  = @response['hits']['hits'].first['_type']
-            raise NoMethodError, "You have tried to eager load the model instances, " +
+            records = {}
+            @response['hits']['hits'].group_by { |item| item['_type'] }.each do |type, items|
+              raise NoMethodError, "You have tried to eager load the model instances, " +
                                  "but Tire cannot find the model class because " +
                                  "document has no _type property." unless type
 
-            begin
-              klass = type.camelize.constantize
-            rescue NameError => e
-              raise NameError, "You have tried to eager load the model instances, but " +
-                               "Tire cannot find the model class '#{type.camelize}' " +
-                               "based on _type '#{type}'.", e.backtrace
+              begin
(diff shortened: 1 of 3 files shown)


karmi:
Hi Andrew, thanks for taking your time to have a look into this. Sadly, this issue has been debated over and over -- please search the issues --, and there are now several patches trying to solve several problems related to this.

The thing is, I feel the codebase right now is too fragile to out it under more load. I like the way you're going in your patch, though, and will definitely look at it when doing some refactorings in February (hopefully).

karmi:
@GearHead90 I like how you approach the problem, Andrew! I generally try to avoid _ActiveSupport_ like a plague, but the `group_by` here works like a charm. I think this is a viable solution. I'll test it as the solution to "multi model searches" and will get back. Please ping me here or on IRC if I'll fail to do that in couple of weeks.

sandrew:
Hello again. I'd like to remind you to make a decision on this request.

karmi:
@sandrew Hi, I like how you approach the problem and will try to incorporate it to the codebase the moment I'll start working on it. If nothing blocks the implementation, I hope we can have it in this week.

allochi:
It would be great if we can have it this week! I'm stuck at this point :D

I have a class I call Resource, and it has 3 children Document, Image and WebPage (and more later), I'm doing all the mapping in Resource class, which is an ActiveRecord class, but it doesn't work, when I force index using rake it just skip the mapping and index all properties.

I wish that this get fixed really soon, so that I don't have to rewrite all my classes again to make Tire work. Please Help :)

Oh, many many thanks for Tire!

I just deleted my comment on this being fixed in the 0.4.0pre, sorry for that, it worked when I set the type to be null in the database, which means no more inheritance.

PR #222, opened 2012-02-01
prefix query

tobowers:
Hi,

Great work on Tire... I'll probably be going through and attempting to add a few more query types and maybe 1.9 support for tests, etc.  Hope this will be ok with you!

Anyway - here's my first one to support the prefix query type.

Thanks,

Topper

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -27,6 +27,14 @@ module Tire
         # TODO: https://github.com/elasticsearch/elasticsearch/wiki/Query-String-Query
         @value
       end
+      
+      def prefix(field, value, options={})
+        if options[:boost]
+          @value = { :prefix => { field => { :prefix => value, :boost => options[:boost] } } }
+        else
+          @value = { :prefix => { field => value } }
+        end
+      end
 
       def custom_score(options={}, &block)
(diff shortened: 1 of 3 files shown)


karmi:
Hi, thanks!

> attempting to add a few more query types

yeah, feel fee to do so.

> and maybe 1.9 support for tests

not sure what you mean by that?

> here's my first one to support the prefix query type.

I'd like to have query types covered by integration tests as well, is it possible to add some?

Best,

Karel

tobowers:
rcov isn't 1.9 compatible so bundle install doesn't work in ruby 1.9+

I don't see any integration tests for other query types... I'm looking in the test/integration directory and I only see query_string.  Am I just not seeing the location?

karmi:
Ah, yes, best to delete the `rcov` then.

See eg. https://github.com/karmi/tire/blob/master/test/integration/range_queries_test.rb etc....

tobowers:
added a integration test

karmi:
@karmi @vhyza Bump!

karmi:
@tobowers Sorry for the ridiculous delay...

lucascaton:
When a new version (with prefix query) will be released?

lucascaton:
?

PR #223, opened 2012-02-01
Only retry Tire::Index#bulk_store on StandardError.

dylanahsmith:
Otherwise bulk_store retries after receiving signals like INT or TERM, likely resulting in the process getting killing by the KILL signal (if being terminated automatically) without executing normal cleanup (e.g. ensure blocks).

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -85,7 +85,7 @@ module Tire
         response = Configuration.client.post("#{Configuration.url}/_bulk", payload.join("\n"))
         raise RuntimeError, "#{response.code} > #{response.body}" if response.failure?
         response
-      rescue Exception => error
+      rescue StandardError => error
         if count < tries
           count += 1
           STDERR.puts "[ERROR] #{error.message}, retrying (#{count})..."
--- a/test/unit/index_test.rb
+++ b/test/unit/index_test.rb
@@ -353,10 +353,24 @@ module Tire
 
         end
 
-        should "try again when an exception occurs" do
+        should "try again when an error response is received" do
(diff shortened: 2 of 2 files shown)


karmi:
Dylan, this would completely change the purpose/semantics of swallowing exceptions from HTTP client etc, wouldn't it?

I think the proper way how to handle this would be to `rescue SystemExit; raise;`...

dylanahsmith:
The HTTP client shouldn't be throwing non StandardError exceptions.  If they are that should be fixed.

karmi:
Yes, seems like you're right, eg. RestClient https://github.com/archiloque/rest-client/blob/master/lib/restclient/exceptions.rb#L82.

Could you try it with RestClient, Curb? Eg. shutdown ES and try the bulk?

Also, the patch will need tests...

dylanahsmith:
Looked at the source code for rest-client and curb, and they both seem to use exceptions that inherit indirectly from StandardError.  I also tried catching the exceptions from `Tire.index('').exists?` when elasticsearch is down with Curb and RestClient, and the exceptions were caught.

Here are some of the inheritance chains:
- Errno::ECONNREFUSED -> SystemCallError -> StandardError
- RestClient::RequestTimeout -> RestClient::RequestFailed -> RestClient::ExceptionWithResponse -> RestClient::Exception -> RuntimeError -> StandardError
- Curl::Err::ConnectionFailedError -> Curl::Err::CurlError -> RuntimeError -> StandardError
- Interrupt -> SignalException -> Exception

I have also added tests to the pull request.

karmi:
Dylan, I'm very sorry for the delay. The patch makes very much sense. Merged & pushed now, thanks!

PR #224, opened 2012-02-02
[BULK] Raise instead of just writing to stderr

nfo:
I happen to have a lot of documents missing, because Tire does not raise in case of failure during a bulk insertions. I think the problem is important enough to be considered as a raised exception.
In my case, my bulk insertions are triggered by Resque jobs. Without an exception, I cannot retry failed jobs.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -92,6 +92,7 @@ module Tire
           retry
         else
           STDERR.puts "[ERROR] Too many exceptions occured, giving up. The HTTP response was: #{error.message}"
+          raise
         end
 
       ensure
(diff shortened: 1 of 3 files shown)


karmi:
Hey, I think the proper way how to handle this would to allow an `options` hash to the `bulk_store` method, which would allow something like `myindex.bulk_store articles, raise: true`...

nfo:
I totally agree. It would need an update of the `import` method too. Do you want me to take care of this ?

karmi:
Yeah, the `options` would have to be passed down the chain to the `import` method in https://github.com/karmi/tire/blob/master/lib/tire/index.rb#L112. Definitely have a shot at that -- just make sure it's covered by unit tests, please!

karmi:
@karmi @vhyza Bump!

karmi:
@nfo Thanks, merged!

PR #226, opened 2012-02-03
Include timeout configuration for ES Requests

bobbytables:
Any word on this?

### DIFF
--- a/lib/tire/configuration.rb
+++ b/lib/tire/configuration.rb
@@ -19,6 +19,10 @@ module Tire
       @logger || nil
     end
 
+    def self.timeout(seconds=nil)
+      @timeout = seconds || @timeout || 10
+    end
+
     def self.reset(*properties)
       reset_variables = properties.empty? ? instance_variables : instance_variables.map { |p| p.to_s} & \
                                                                  properties.map         { |p| "@#{p}" }
@@ -26,5 +30,4 @@ module Tire
     end
 
   end
-
 end
--- a/lib/tire/http/client.rb
+++ b/lib/tire/http/client.rb
@@ -8,7 +8,7 @@ module Tire
         ConnectionExceptions = [::RestClient::ServerBrokeConnection, ::RestClient::RequestTimeout]
 
         def self.get(url, data=nil)
-          perform ::RestClient::Request.new(:method => :get, :url => url, :timeout => Configuration.timeout, :payload => data).execute
+          perform ::RestClient::Request.new(:method => :get, :url => url, :timeout => Configuration.timeout, :open_timeout => Configuration.timeout, :payload => data).execute
         rescue *ConnectionExceptions
           raise
         rescue ::RestClient::Exception => e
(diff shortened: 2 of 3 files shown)


karmi:
> Any word on this?

Yeah :)
- Why submit the issue as a new one and not ping me on the original one?
- Care to include some explanation what this does and why you need it? I'm not sure what exact problem are you solving by reading the code.
- I can see the timeout is propagated to the RestClient only? What about other clients?
- What about wrapping the calls in Ruby's [`timeout`](http://www.ruby-doc.org/stdlib-1.9.3/libdoc/timeout/rdoc/Timeout.html)? Is there something to gain/lose?
- What about the elasticsearch's own `timeout` parameter? Is it of any use here?

bobbytables:
Sweet.
1. Because my original one got an extra commit that I pushed to my branch.
2. We're running ES on EC2 and our stack is on Rackspace so we wanted to add timeouts incase anything exceeded 1 second. (Our EC2 instances are working hard on something else, etc...)
3. It was added there because those were the only places I could see calls being made. Are there others?
4. RestClient supports open_timeout and timeout. In the future it would be easy to support both, for now this covers both of them with one configuration. I don't see a benefit to using Ruby's Timeout for the simple reason that it gives no additional functionality to the end purpose.
5. Timeouts in ElasticSearch could produce unexpected results if worded this way. They do 2 different things. It could, however, be a part of the query DSL.

dylanahsmith:
> What about wrapping the calls in Ruby's timeout? Is there something to gain/lose?

This wouldn't allow the rest-client timeouts to be longer than the defaults.  For instance, `curl localhost:9200/#{index_name}/_optimize?max_segments=5&refresh=true&flush=true` could easily take over the default read_timeout of 60 seconds after reindexing a large index.  It is also a case where increasing the read_timeout but not having a long open_timeout makes sense.  It would be nice to be able to do this with tire's http client.

bobbytables:
There might be good reason to have read_timeout and open_timeout as separate configurations. Perhaps even keep the simple "timeout" configuration that does both for you.

karmi:
I think the more appropriate solution to this cluster of problems would be to do something I've planned for a long time -- use [_Faraday_](https://github.com/technoweenie/faraday) as the HTTP client abstraction and let it handle all of this?

bobbytables:
I could get behind that. But what more does Tire do in terms of HTTP request outside of the http/client class?

bobbytables:
@karmi ?

karmi:
@bobbytables I still don't have any strong opinion on this, other then intuitively I don't think this is a proper & future-proof approach.... I need to mull it over a bit more, I guess, sorry.

jedi4ever:
Yeah, why not just go in steps? Include the timeout to get people going, adding another Faraday is not going to be harder by doing so?

karmi:
Allright, _really_ sorry about the delay... The patch only adds the behaviour to RestClient, where most people (should) use Curb in production, the semantics are weird because the timeout is global (I know, my fault by designing Configuration this way), ... I'm sorry, leaning towards passing on this, in favour or proper implementation...

karmi:
@jedi4ever Hard to explain. Aesthetics and code as liability.

To elaborate, I don't think this is a proper solution to the problem, and after merging this _I'm_ the one taking care of the code and possibly issues arising from it... I know you're maintaining a larger code base -- this is just my way of keeping myself sane, I'm afraid...

karmi:
Closing in favour of a more complete implementation.

jedi4ever:
@karmi thanks for the comments. looking forward to new implementation . :+1:

PR #227, opened 2012-02-04
Documentation: Global facet option

namxam:
Hey, it is just a very small fix, but the documentation is wrong about where to put the ":global => true" option.


karmi:
Ah, yes, thanks!

Issue #84, opened 2011-08-30
While importing, percentage complete exceeds 100%


jpsilvashy:
Upon running import for my model like so:

```
rake environment tire:import CLASS='Recipe.all' INDEX='mongo-recipes' FORCE=1 RAILS_ENV='production'
```

I receive some strange behavior resulting in a failure to import the model completely. 

```
** Invoke environment (first_time)
** Execute environment
** Invoke tire:import (first_time)
** Execute tire:import
#2011-08-30 08:00:05:130 [DELETE] ("mongo-recipes")
#
curl -X DELETE "http://localhost:9200/mongo-recipes"

#2011-08-30 08:00:05:130 [200]

#2011-08-30 08:00:05:173 [CREATE] ("mongo-recipes")
#
curl -X POST "http://localhost:9200/mongo-recipes" -d '{"mappings":{"recipe":{"properties":{"title":{"type":"string","analyzer":"snowball","boost":100},"description":{"type":"string","analyzer":"snowball","include_in_all":false},"tags":{"type":"array"},"id":{"type":"string"}}}}}'

#2011-08-30 08:00:05:173 [200]

[IMPORT] Deleting index 'mongo-recipes'
[IMPORT] Creating index 'mongo-recipes' with mapping:
{"recipe":{"properties":{"title":{"type":"string","analyzer":"snowball","boost":100},"description":{"type":"string","analyzer":"snowball","include_in_all":false},"tags":{"type":"array"},"id":{"type":"string"}}}}
[IMPORT] Starting import for the 'Recipe.all' class
--------------------------------------------------------------------------------
26733/26733 | 100% # 2011-08-30 08:00:07:840 [BULK] ("mongo-recipes")
#
curl -X POST "http://localhost:9200/_bulk" -d '{... data omitted ...}'

#2011-08-30 08:00:07:841 [200]

53466/26733 | 200% # 2011-08-30 08:00:10:872 [BULK] ("mongo-recipes")#######################################################
#
curl -X POST "http://localhost:9200/_bulk" -d '{... data omitted ...}'

#2011-08-30 08:00:10:872 [200]

80199/26733 | 300% # 2011-08-30 08:00:14:477 [BULK] ("mongo-recipes")#####################################################################################################################
#
curl -X POST "http://localhost:9200/_bulk" -d '{... data omitted ...}'
```

Eventually this grows to about 2400% and dies with:

```
curl -X POST "http://localhost:9200/_bulk" -d '{... data omitted ...}'

[...]

PR #232, opened 2012-02-06
[#84] Fixed showing the percentage incorrectly when import from Mongoid ...

stnguyen:
...(tested with mongoid 2.4.1)


karmi:
@stnguyen Sorry for the long delay, Stefan, merged and pushed, thanks!

PR #238, opened 2012-02-10
Avoiding conflict with meta_search (or other gems) when it overrides the search method.

goncalvesjoao:
Hi there karmi and vhyza!

First of all, thanks for this wonderful gem.
I've been using it in a project that also uses the "meta_search" gem, but this one overrides the search method that Tire declares.

I've simply rename the tire search method to "search_with_tire" and declared the search method like so:
        def search(_args, &block)
          search_with_tire(_args, &block)
        end
This keeps the standard behavior intact and allows people like me to acces the tire search method with an alternative call: "search_with_tire".

The changes are fairly simple, could you guys incorporate them in your master version?
I prefer to use your gem instead of my forked one.

Cheers.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -63,6 +63,10 @@ module Tire
         #     Article.search :load => { :include => 'comments' } do ... end
         #
         def search(*args, &block)
+          search_with_tire(*args, &block)
+        end
+        
+        def search_with_tire(*args, &block)
           default_options = {:type => document_type, :index => index.name}
 
           if block_given?


karmi:
Hi,

the behaviour you're describing is weird -- _Tire_ goes to great lengths to _not_ redefine existing methods on your models. Of course, you need to `include` it as last. Please check the Readme for detailed information on this.

If the problem persist, please post an example of your model code (or a failing test)!

karmi:
@MacFrancis Closing this for the moment, please submit another issue demonstrating the problem -- Tire should _not_ by any means trample on any existing methods. Again, make sure you `include` everything in proper order.

PR #242, opened 2012-02-14
Support for ElasticSearch's 'explain' => true in query and '_explanation' field in the resulting items

redbeard:
This commit includes unit tests, an integration test and an implementation for turning on the explanation field with the scoring breakdown. 

This is useful when developing user-interfaces for building and exploring search queries.

Usage:

``` ruby
        s = Tire.search('explanation-test') do
          query { string 'content:Fox' }
          explain true
        end
```

Which enables the '_explanation' attribute on each result item, accessible so:

``` ruby
s.results.first._explanation
```

### DIFF
--- a/lib/tire/model/persistence.rb
+++ b/lib/tire/model/persistence.rb
@@ -45,7 +45,7 @@ module Tire
 
           include Persistence::Storage
 
-          ['_score', '_type', '_index', '_version', 'sort', 'highlight', 'matches'].each do |attr|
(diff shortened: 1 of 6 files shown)


karmi:
@karmi @vhyza Bump!

thoughtpunch:
What happened with this? This was an awesome feature!

karmi:
@thoughtpunch It was closed by karmi/tire@24b3931

heberuriegas:
Do you know if is possible call ._explanation with the "load: true" in the search function?

PR #247, opened 2012-02-20
Bulk 'create' indexing

haarts:
Hi,

I have extended the bulk_store method to accept an optional hash argument. When the options hash contains a :method key the value of this key is used to signal the kind of bulk indexing which should be done ('create' in this instance).

I included a test in it's own separate file as I was unsure where to include it in the existing test library (probably in test/unit/index.rb).

This feature is build on the latest master and should apply cleanly.

With kind regards,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.un~
 *.gem
 .bundle
 Gemfile.lock
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -68,18 +68,9 @@ module Tire
       options.merge!({:method => "index"}) unless options[:method]
 
       payload = documents.map do |document|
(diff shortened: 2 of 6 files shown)


haarts:
We changed `bulk_store` method some more in order to support parent/child relations in the bulk import functionality. We extracted the string interpolated header line into a separate method for clarity.

woahdae:
+1 for parent-child.

karmi:
Hey, I went with a slightly lighter but more complete implementation: https://github.com/karmi/tire/compare/4e85109...1c0d043

Could you test it out?

woahdae:
left my thoughts [here](https://github.com/karmi/tire/pull/449#issuecomment-10431465)

PR #248, opened 2012-02-20
Support for simple wildcard queries

tieleman:
I've added support for the wildcard query type. We use this quite a bit, might be interesting for you as well.

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -21,6 +21,10 @@ module Tire
         @value = { :range => { field => value } }
       end
 
+      def wildcard(field, value)
+        @value = { :wildcard => { field => value } }
+      end
+      
       def text(field, value, options={})
         query_options = { :query => value }.update(options)
         @value = { :text => { field => query_options } }
--- a/test/integration/wildcard_queries_test.rb
+++ b/test/integration/wildcard_queries_test.rb
@@ -1,0 +1,25 @@
+require 'test_helper'
+
+module Tire
+
+  class WildcardQueriesIntegrationTest < Test::Unit::TestCase
+    include Test::Integration
+
+    context "Wildcard queries" do
+
+      should "allow simple wildcard queries" do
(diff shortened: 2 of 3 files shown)


niuage:
So, is `wildcard "title" "f*"` just a nicer way to write `string "title:f*"` or is there a difference?

tieleman:
It's a different type of query in elasticsearch, see:

http://www.elasticsearch.org/guide/reference/query-dsl/wildcard-query.html

Allows for stuff like: `{ wildcard  => { tags => 'm*g?c' } }`

karmi:
Hi, I'm asking the same thing as @niuage -- _why exactly_ have a different query when everybody can do `string "title:f*"`?

(Of course, we know it's a different ES query, but then ES have lots of different queries and I'm not sure we should support them all :)

That aside, I think this would make a terrific _Tire_ extension in [**tire-contrib**](https://github.com/karmi/tire-contrib)! I see _tire-contrib_ as a lively place, where we can stash lots of stuff, people can poke around, all without putting a burden on the core library.

In this case, the user would probably do something like:

``` ruby
require 'tire'
require 'tire/queries/wildcard'

s = Tire.search 'foobars' do
  query { wildcard 'foo?ar' }
end
```

It means it would be implemented in file `tire-contrib/lib/tire/queries/wildcard.rb`. What do you think?

I'll close the pull request here, so we can continue on _tire-contrib_, is that OK?

tieleman:
Very well, I'll see if I can cook up a patch.

cjbottaro:
Yes, there is a difference.

Does _not_ work:

```
f.query{ string("form_field_10061:*UNT*") }
```

Does work (using this patch):

```
f.query{ wildcard("form_field_10061", "*UNT*") }
```

`form_field_10061` is a `not_analyzed` `string` field.  Using Elasticsearch-0.19.11.

I vote to accept this pull request... :P

PR #249, opened 2012-02-20
Added default_search_filter options. Similar to 'default_scope' of Activerecord

siliconsenthil:
Hi,

I have added a commit to support default-search-filter.

When declared, it adds a term filter with the given term hash to all searches. If user wishes to search excluding this filter options[:unscoped] has to be passed as true (similar to AR).

I've added integration and unit tests as well.

Feel free to give comments. I can add/remove/modify based on that.

Thanks.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -85,6 +85,8 @@ module Tire
             end
           end unless sort.empty?
 
+          s.filter(:term, klass.default_filter)  if klass.default_filter && !options[:unscoped]
+
           if block_given?
             block.arity < 1 ? s.instance_eval(&block) : block.call(s)
           else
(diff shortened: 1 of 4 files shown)


karmi:
Hi, it's an interesting idea!

So, let's review the _ActiveRecord_ feature here: http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Default/ClassMethods.html#method-i-default_scope. I can see how that could make sense with _Tire_.

First, I am 100% sure this would make a terrific [**tire-contrib**](https://github.com/karmi/tire-contrib) contribution. Maybe I'm not pushing _tire-contrib_ enough, maybe people struggle with how to extend _Tire_ with the contrib, but that's a place where features like this should be proposed -- and quite possibly implemented.

Second, if I understand your [code](https://github.com/siliconsenthil/tire/blob/default_filter/lib/tire/model/search.rb#L88) properly:

``` ruby
s.filter(:term, klass.default_filter)  if klass.default_filter && !options[:unscoped]
```

you just add a `term` filter to the search block. This feels to me very limited compared to the rich semantics of _ActiveRecord_'s `default_scope`. I think we should strive for something similar, along the lines of:

``` ruby
class ArticleWithDefaultFilter < ActiveRecord::Base
  include Tire::Model::Search

  mapping do
    indexes :status,  type: 'string', index: 'not_analyzed'
    indexes :user_id, type: 'string', index: 'not_analyzed'
  end

  default_search_filter do |f|
    f.term  :status, 'active'
    f.term  :user_id, current_user.id
  end
end
```

Exposing the _full_ _Tire's_ DSL in the "default scope filter" like this would make more sense?

Please feel free to submit it as a [tire-contrib](https://github.com/karmi/tire-contrib) pull request, I'll close this one. It would be great if anybody would be willing to help @siliconsenthil ?

PR #255, opened 2012-02-26
FilteredQuery: chain multiple filters with and by default

### LINKED ISSUE #254: Only last filter in FilteredQuery used
Tire is using an array within the "filter" value of filtered queries. The [Filtered Query elasticsearch documentation](http://www.elasticsearch.org/guide/reference/query-dsl/filtered-query.html) uses a hash, so this is not a documented feature of elasticsearch.  I found that only the last filter in the filter array will be used by elasticsearch for this incorrect syntax.

Running the following example shows the problem.

``` ruby
#!/usr/bin/env ruby

require 'tire'

Tire.index 'bug_test' do
  delete
  create
  store :title => 'One',   :tags => ['ruby'],           :published => false
  store :title => 'Two',   :tags => ['ruby', 'python'], :published => true
  store :title => 'Three', :tags => ['java'],           :published => true
  store :title => 'Four',  :tags => ['ruby', 'php'],    :published => false
  refresh
end

Tire.configure { logger STDOUT, :level => 'debug' }

s = Tire.search 'bug_test' do
  query do
    filtered do
      query { all }
      filter :term, :published => true
      filter :term, :tags => 'ruby'
    end
  end
end

s = Tire.search 'bug_test' do
  query do
    filtered do
      query { all }
      filter :term, :published => true
      filter :term, :tags => 'ruby'

      # Hack to wrap the filters array in an :and filter
      to_hash[:filter] = { :and => to_hash[:filter] }
    end
  end
end
```

Two searches are made, and logged to STDOUT.

The first search (body shown below) returns 3 documents, including the ones with published => false, which should be filtered. 

``` json
{
    "query": {
        "filtered": {
            "query": {
                "match_all": {}
            },
            "filter": [{
                "term": {
                    "published": true
                }
            }, {
                "term": {
                    "tags": "ruby"
                }
            }]
        }
    }
}
```

The second search (body shown below) wraps the filter array within an :and filter, so both filters apply and one results is returned.

``` json
{
    "query": {
        "filtered": {
            "query": {
                "match_all": {}
            },
            "filter": {
                "and": [{
                    "term": {
                        "published": true
                    }
                }, {
                    "term": {
                        "tags": "ruby"
                    }
                }]
            }
        }
    }
}
```

ralph:
Fixes #254

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -127,8 +127,9 @@ module Tire
       end
 
       def filter(type, *options)
-        @value[:filter] ||= []
-        @value[:filter] << Filter.new(type, *options).to_hash
+        @value[:filter] ||= {}
+        @value[:filter][:and] ||= []
+        @value[:filter][:and] << Filter.new(type, *options).to_hash
         @value
       end
 
--- a/test/integration/filtered_queries_test.rb
+++ b/test/integration/filtered_queries_test.rb
@@ -23,15 +23,32 @@ module Tire
         assert_equal 'Two', s.results.first.title
       end
 
-      should "restrict the results with multiple filters" do
+      should "allow other filters filters" do
         # 2.json > Is tagged "ruby" and has 250 words
 
         s = Tire.search('articles-test') do
           query do
             filtered do
               query { all }
-              filter :and, { :terms => { :tags => ['ruby', 'python'] } },
-                           { :range => { :words => { :from => '250', :to => '250' } } }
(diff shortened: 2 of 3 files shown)


karmi:
Thanks @ralph, merged and pushed.

One note: could be fun, sometimes in the future, if the `filtered` method (and the corresponding `FilteredQuery` class) took some `options`, where the user could specify if she wants `and` or `or` chain?

PR #256, opened 2012-02-28
Tire::Index#store supports the "parent" option

nickhoffman:
Hi there, Karel. This pull request enables users to store documents in a parent-child relationship.

`Tire::Index#store` can be told to set the `parent` option in the query string. Eg:

``` ruby
Product.tire.index.store doc, :parent => doc.parent.id
```

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -48,13 +48,20 @@ module Tire
       if options
         percolate = options[:percolate]
         percolate = "*" if percolate === true
+        parent = options[:parent]
       end
 
       id       = get_id_from_document(document)
       document = convert_document_to_json(document)
 
       url  = id ? "#{Configuration.url}/#{@name}/#{type}/#{id}" : "#{Configuration.url}/#{@name}/#{type}/"
-      url += "?percolate=#{percolate}" if percolate
+
+      if percolate || parent
+        query_string = []
+        query_string << "percolate=#{percolate}" if percolate
+        query_string << "parent=#{parent}" if parent
+        url += "?" + query_string.join("&")
+      end
 
       @response = Configuration.client.post url, document
       MultiJson.decode(@response.body)
--- a/test/unit/index_test.rb
+++ b/test/unit/index_test.rb
@@ -178,6 +178,15 @@ module Tire
           @index.store :title => 'Test'
         end
 
+        context "with the :parent option" do
+
+          should "set the :parent option in the query string" do
+            Configuration.client.expects(:post).with("#{Configuration.url}/dummy/document/?parent=1234", '{"title":"Test"}').returns(mock_response('{"ok":true,"_id":"test"}'))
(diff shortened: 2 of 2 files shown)


karmi:
Hi Nick, as with #258, this is all hacking what should be done more systematically, exposing all the useful parameters for Search API. In the end, all of this is just a simple hash of options, encoded as URL params.

I'm reluctant to proceed with that in a piecemeal fashion...

nickhoffman:
I hear you! It'd be great to support all of these options in one nice refactoring, rather than, as you said, adding support for options one by one.

Do you have any ideas on how to support all of the options for the ES' Search API?

karmi:
Yes, all the options are scattered around, see eg. [`model/search.rb`](https://github.com/karmi/tire/blob/master/lib/tire/model/search.rb#L92-94) which is crazy and hell from maintenance standpoint.

For search, I think we should somehow work with possible options directly in [`lib/tire/search.rb`](https://github.com/karmi/tire/blob/master/lib/tire/search.rb#L12)! I must to have a look into that...

Maybe we can just serialize options send to `Index#store` and pass them to ES? Once I'm done with all the pull requests and stuff, I'll try to have a look into that as well... But feel free to submit a pull request for that, Nick!

nickhoffman:
Serializing the options to [`Tire::Index#store`](https://github.com/karmi/tire/blob/master/lib/tire/index.rb#L44-65) might be the fastest and simplest approach. If I submit a pull request for that, it'd be merged in?

mattiassvedhem:
Is there a work around for doing this ?

mattiassvedhem:
Hi @karmi any update on this? What could be done to move in this direction?

karmi:
@yeggeps Making sure `Index#store` works with the passed in `options`, calling `to_param` on them as eg. https://github.com/karmi/tire/blob/master/lib/tire/search.rb#L119, not breaking anything (= not sending bogus options down to ES), proper unit tests for the feature and ideally an integration test demonstrating the _purpose_ (not the bare functionality) of the feature...

mattiassvedhem:
@karmi cool, has any work been done on this? if not I might take a stab at it.

karmi:
@yeggeps Not yet. I would definitely try to solve all the `options` related issues across the whole codebase as soon as possible.

karmi:
@nickhoffman Nick, sorry about taking so long to process this... In the end, I've cleaned up the params handling in `Index#store` opening it up for support for other URL parameters, such as routing, ttl, etc.

mattiassvedhem:
Awesome @karmi! Any idea when it will make it to the gem? Thanks.

karmi:
@mattiassvedhem No idea :), please use the Git URL in the `Gemfile` for the time being...

nickhoffman:
Great work, @karmi. Thanks for not forgetting about this feature.

Issue #88, opened 2011-09-01
Add ability to specify the routing of a document


changa:
It would be great if we could specify the [routing](http://www.elasticsearch.org/guide/reference/api/search/index.html) when indexing a Tire::Model.

karmi:
Yeah. How should the interface look like? Options for the `search` method or part of the DSL?

(BTW, I think it would be much nicer to have the [`search_type`](http://www.elasticsearch.org/guide/reference/api/search/request-body.html) exposed.)

changa:
Hi (and sorry for the lag) !
I guess it would be fine as options to `search`.
Concerning `search_type`, you're right, it certainly could prove useful.

(And BTW, thanks a lot for this gem, it rocks :-)

benzheren:
+1 for this feature, it will definitely helps the performance of the search itself.

PR #258, opened 2012-02-29
Add support for routing parameter to search requests.

dylanahsmith:
Addresses specifying the routing parameter for search. See issue #88.

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -12,6 +12,7 @@ module Tire
         @options = options
 
         @path    = ['/', @indices.join(','), @types.join(','), '_search'].compact.join('/').squeeze('/')
+        @params = options.slice(:routing)
 
         block.arity < 1 ? instance_eval(&block) : block.call(self) if block_given?
       end
@@ -25,7 +26,9 @@ module Tire
       end
 
       def url
-        Configuration.url + @path
+        query = @params.to_param
+        query = "?" + query unless query.empty?
+        Configuration.url + @path + query
       end
 
       def query(&block)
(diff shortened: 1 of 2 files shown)


karmi:
Hi Dylan, I don't know, this seems to me as another case of trying to hack some ES options into Tire methods. I've said it previously, we need to expose [all the options](http://www.elasticsearch.org/guide/reference/api/search/request-body.html), namely `search_type` etc somehow. I'm reluctant to do it in piecemeal fashion.

dylanahsmith:
>  we need to expose [all the options](http://www.elasticsearch.org/guide/reference/api/search/request-body.html), namely `search_type`

The difference is that those are body parameters, but this one is a URL parameter and should be a place that is easier to share with the Tire.search(indices, payload) code path of Tire.search. See pull request #259 for how I made this usable for both code paths of Tire.search.

> this seems to me as another case of trying to hack some ES options into Tire methods

I don't see how not implementing unrelated functionality somehow makes this code a hack.  It certainly doesn't make it harder to implement support for search body parameters.

I also implemented this in a method that allows that makes it easier to add additional URL parameters (although the only other I can find is `pretty`) by adding them to the `slice` method, and it is implemented in a way that allows a methods in `Tire::Search::Search` to set a query parameter by modifying `@params`.

If you want, it would be very easy to modify this code to only expose the routing parameter through a method in `Tire::Search::Search`.  But so far you haven't expressed any opinion either way in issue #88, so I thought I would try to get some feedback.

**Edit:** Looked at `search_type` more carefully, and it does appear to be a url parameter as well, even though it was listed along with body parameters in the docs.

erickt:
@dylanahsmith: `search_type` is only a url parameter, it cannot be passed in the body.

@karmi: This approach seems fine to me. I'd be fine on punting on #198 for this.

PR #259, opened 2012-02-29
Fix Tire.search with a payload rather than a block.

dylanahsmith:
Previously calling Tire.search with a string payload would cause an exception to be raised in Tire::Results::Collection, since the payload was passed directly into the Collection.new options parameter.

Passing any options to Tire.search would result in those options being encoded directly in the body of the search, causing the search to fail.

This also resulted in missing functionality for Tire.search with a payload, since options couldn't be specified, such as the document type. This commit adds support for this through specifying the search body in a payload option, then recognizing additional options supported by Tire.search with a block.

### DIFF
--- a/lib/tire/dsl.rb
+++ b/lib/tire/dsl.rb
@@ -20,7 +20,7 @@ module Tire
         end
 
         search = Search::Search.new(indices, options) { |search| nil }
-        response = Configuration.client.post search.url, payload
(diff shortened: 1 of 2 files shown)


dylanahsmith:
The issues this pull request was addressing have been addressed by commit e3e09a6b99565610b67d89a33cdad0afb2222203.  Thanks!

PR #262, opened 2012-03-05
add test showing highlighting with more than one field

phoet:


plentz:
+1

karmi:
All right, thanks! But, why not just:

```
highlight :tags, :title, :options => { :tag => '<strong>' }
```

phoet:
simple test is simple ;)

PR #265, opened 2012-03-07
fixed #264 "You have tried to eager load the model instances..."

### LINKED ISSUE #264: You have tried to eager load the model instances, but Tire cannot find the model class
Tire::Results::Collection tries to infer class name when using `load: true`, but cannot do so if it's for a namespaced class.

For instance OnlineStore::Product results in _type == online_store_product, of which we cannot convert back to the appropriate constant.

I would like to add the option for specifying the classname when doing a `load: true`

sarmiena:
This allows us to specify an explicit model name when calling tire.search or from the DSL itself
e.g. tire.search(load: true, class_name: 'My::ClassName') { ... }

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -47,9 +47,9 @@ module Tire
 
             begin
               explicit_class_name = @options[:class_name]
-              klass = explicit_class_name.nil? ? type.camelize.constantize : @options[:class_name].constantize
+              klass = explicit_class_name.nil? ? type.camelize.constantize : explicit_class_name.constantize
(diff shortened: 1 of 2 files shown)


demersus:
This would be nice to have in the main gem!  I will need to use this patch immediately in my app with namespaced models.

karmi:
@sarmiena @demersus Should be closed in d5c08fb, can you check, please?

sarmiena:
I'll take a look when I get home tonight

karmi:
Closing due to long ping latency :)

PR #267, opened 2012-03-07
Don't use to_sym in Tire::Results::Item, as keys from elastic search might provide from user input

asterite:
Hi,

I see that Tire::Results::Item calls to_sym on every key-value pair retrieved from ElasticSearch.

I have an app where a user has sites on a map, and every site can have a bunch of user-defined properties. Those properties with their values get stored in ElasticSearch.

When those properties are retrieved from ElasticSearch they are first converted to Symbols. **This is a memory leak, because Symbols are never released. So, a malicious user could potentially generate sites with randomly generated user-defined properties, and so the memory will be filled of Symbols.**

In general, it's never good to use to_sym from user generated data.

The commit I made just used to_s instead. Since ElasticSearch already returns the json keys as strings, calling to_s just returns the string itself. When indexing by a symbol, I also invoke to_s, so that would be the only "performance problem" (but later that string can be freed).

### DIFF
--- a/lib/tire/results/item.rb
+++ b/lib/tire/results/item.rb
@@ -13,9 +13,9 @@ module Tire
         @attributes = {}
         args.each_pair do |key, value|
           if value.is_a?(Array)
-            @attributes[key.to_sym] = value.map { |item| @attributes[key.to_sym] = item.is_a?(Hash) ? Item.new(item.to_hash) : item }
+            @attributes[key.to_s] = value.map { |item| @attributes[key.to_s] = item.is_a?(Hash) ? Item.new(item.to_hash) : item }
           else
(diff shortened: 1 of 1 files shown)


karmi:
This still feels too esoteric to me... I understand the [issue](http://www.tricksonrails.com/2010/06/avoid-memory-leaks-in-ruby-rails-code-and-protect-against-denial-of-service/), but don't know if having hundreds of new Strings floating around helps memory wise... Thanks!, but I think I'll pass on this one...

karmi:
Seems like `String` is note safe either, http://www.ruby-lang.org/en/news/2012/11/09/ruby19-hashdos-cve-2012-5371/

PR #268, opened 2012-03-07
Add ability to override to_param for Results::Item

willvr:
I wanted to be able to send a Results::Item directly to a named route like

```
  video_path(result)
```

for a model with to_param redefined to a "friendly string".  I wanted to ability to specify a to_param attribute.  I'm not sure if there's a preferred way to do this, but this patch works for me.

### DIFF
--- a/lib/tire/results/item.rb
+++ b/lib/tire/results/item.rb
@@ -51,6 +51,10 @@ module Tire
         persisted? ? [id] : nil
       end
 
+      def to_param
+        @attributes[:to_param] || [id] if persisted?
+      end
+
       def to_hash
         @attributes
       end


karmi:
@will-kidzui Hi, `Results::Item` should be fully "Rails compatible", especially in routes. See the generated application. Could you provide an example of code where it does not work?

Check that `to_param` [depends](https://github.com/rails/rails/blob/dc8773b19f61af2ba818d66923fc65e17bad6c20/activemodel/lib/active_model/conversion.rb#L51) on `to_key`, which is [indeed](https://github.com/karmi/tire/blob/master/lib/tire/results/item.rb#L50-52) implemented.

andruby:
Hi @karmi,

I have the same challenge as @willvr: one of my models defines `to_param` as a unique string (aka: a permalink).

``` ruby
  include Tire::Model::Search
  def to_param
    self.permalink
  end
end
```

When I call `url_for()` on an `Article` instance it returns the correct url, but on a `Results::Item` it will use the id instead of the permalink:

``` ruby
url_for(Article.first) # => /articles/hello-world-blogpost
url_for(Article.search('*').results.first) # => /articles/1
```

The code added in @willvr's pull request would allow me to add the `to_param` attributes in the `to_indexed_json` method.

``` ruby
def to_indexes_json
  to_json(methods: [:to_param])
end
```

Is there a way to solve our challenge with the Tire code as is, or do we need something like this pull request?

Cheers,
Andrew

tamird:
@karmi this is still an issue exactly as described by @andruby. Please consider reopening this for discussion, at least.

tamird:
Currently the only workaround:

``` Ruby
article = Article.search('*').results.first

def to_param(object)
  <copy your implementation from model>
end

article_path(to_param(article))
```

blarralde:
That'd be useful for me too. I've got results from multiple models and it's much cleaner if I can just do `link_to result.name, result`

karmi:
@tamird @blarralde Hi, reopening, will try to emulate & test it on my end.

mattiassvedhem:
You could consider implementing `to_param` in a shared [Draper decorator](https://github.com/karmi/tire/wiki/Using-Tire-with-Draper). That way you can reuse the implementation between the Item and the model class. Also, you'll get a bonus, since it's a more appropriate layer to implement it at anyways, arguably.

Of course, you would still have to pass any attributes needed to construct it, down to ElasticSearch.

martijn:
This is something I run into as well.. another thing I notice with my Mongoid models is that a specific Time field is returned as a string. The results I get from tire look like my regular Mongoid instances, but they seem quite different.

mattiassvedhem:
@martijn You probably need to specify the date mapping for the given field.

The objects you are interacting with is instances of [Item](https://github.com/karmi/tire/blob/master/lib/tire/results/item.rb) It's not the underlying Mongoid Model, unless you specify `:load => true`.

martijn:
Ah that was not immediately obvious to me, since:

```
Foo.tire.search('bar').first.class # => Foo
```

Thanks!

mattiassvedhem:
@martijn Yep, in Rails, Item instances [masquerades](https://github.com/karmi/tire/blob/master/lib/tire/results/item.rb#L80) as the model.

PR #274, opened 2012-03-11
More Like This Queries

mereghost:
Hey karmi & vhyza,

I've implemented the more like this queries, if anything else is needed for the pull request to be accepted let me know.

Thanks in advance

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ examples/*.html
 .rvmrc
 .rbenv-version
 tags
+.idea
--- a/lib/tire.rb
+++ b/lib/tire.rb
@@ -41,3 +41,4 @@ module Tire
   end
   module_function :warn
 end
+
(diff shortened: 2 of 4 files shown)


karmi:
Hi, thanks!

First of all, could you transfer the code to a pull request on [`tire-contrib`](https://github.com/karmi/tire-contrib)? I'd like to keep the _Tire's_ core as slim as possible, and even rip out some things such as `Tire::Persistence` to _contrib_.

Have a look at the [`rails-logger`](https://github.com/karmi/tire-contrib/blob/master/lib/tire/rails-logger.rb) extension. You can provide README for the extension, tests, etc.

Second, I think the `mlt` does not fit well into the DSL, what about `more_like_this`?

Also, could you please add some integration tests as well -- they are generally the best "documentation" for the code!

mereghost:
Okay. I'll do that and re-send the pull to tire-contrib. =)

PR #278, opened 2012-03-14
Implemented auto reindex when associated model is updated

chromeragnarok:
Hi all,

I've implemented a feature which trigger a reindex when an associated model is updated. For example:

```
    class ::AssociatedModel < ActiveRecord::Base
      include Tire::Model::Search

      has_many :articles,    :class_name => "ActiveModelArticleWithAssociation",    :foreign_key => "article_id"

      mapping do
        indexes :first_name
        indexes :last_name
      end
    end

    class ::ActiveModelArticleWithAssociation < ActiveRecord::Base
      include Tire::Model::Search

      mapping do
        indexes :title
        indexes :content
        indexes :associated_model, :class => AssociatedModel do
          indexes :first_name
        end
      end
    end
```

Everytime an instance of AssociatedModel first_name is updated, any referring instance of ActiveModelArticleWithAssociation will be reindexed. Thus keeping the index in sync with the database. The indexing is (preferrably) done through delayed_job. If there's no delayed_job available, it will be reindex on the main application.

Second minor feature that I've added is that nested attribute will be created automatically, based on the mapping. So no need to define a separate to_indexed_json in your class. However, overriding the to_indexed_json would still work as well. At the moment, it only support 1 level of nesting, I don't think it will support any deeper levels at the moment.

Thanks!


chromeragnarok:
Oh and one more thing, to enable the auto reindex you have to add nested_attributes into your Tire.configure:

```
Tire.configure {
  nested_attributes :delayed_job => true do
    nest :associated_model => :active_model_article_with_association
  end
}
```

The delayed_job option is used if you want to use delayed_job, default is false. For multiple class you just add another nest inside the block:

```
Tire.configure {
  nested_attributes :delayed_job => true do
    nest :associated_model => :active_model_article_with_association
    nest :associated_model => :another_active_model_article_with_association
  end
}
```

If nested_attributes is not added, the auto reindex will be disabled.

plentz:
-1 because of the delayed_job dependency that it add(but I liked the idea itself)

chromeragnarok:
Ah right, the reasoning was that, I didn't want to add more load to the main application thread when it does the automatic reindexing. Thus, doing it on delayed_job is much more preferable. You could set the delayed_job option to false and let it handle the auto reindexing on the main application thread.

donaldpiret:
Could you detect the presence of Delayed job (By checking wether the delayed module is defined) and just use it if it is by default? 
That would make the delayed_job option optional

chromeragnarok:
Yup, its doing that here : https://github.com/chromeragnarok/tire/blob/master/lib/tire/configuration.rb#L30

However, delayed_job is still needed as the gem development dependency.

.gitignore
```diff
@@ -6,6 +6,7 @@ rdoc/
 doc/
 .yardoc/
 coverage/
+.idea/
```
ches, line 9: General pet peeve: please use a global [`excludesfile`](https://help.github.com/articles/ignoring-files) for your IDE's litter so everyone's projects don't need to know about your personal choice of tools.

karmi, line 9: :+1:  :)

karmi:
Hi, I'm too busy to understand what exact problem are you trying to solve here. These are two separate documents in the index? Because for common cases, you just [`touch`](http://stackoverflow.com/questions/11692560/elasticsearch-tire-and-nested-queries-associations-with-activerecord/11711477#11711477) the associated record.

If you clean the code up and add some documentation with proper examples, I think it's a nice contribution to karmi/tire-contrib...

chromeragnarok:
Great thanks for pointing me out to that touch article. If I have the extra time, I'll try to add more documentation and cleanup the code, and will try to resubmit again :)

Issue #275, opened 2012-03-11
Explictly specifying type 


philly-mac:
Hi,

I don't know if it is just me, but I find it kind of annoying and even irregular that in the tire code you can specify the type like this when removing from the index

``` ruby
def remove(*args)
      if args.size > 1
        type, document = args
        id = get_id_from_document(document) || document
      else
        document = args.pop
        type = get_type_from_document(document)
        id = get_id_from_document(document) || document
      end
```

but when you are adding to the index you have to place the type within the structure of the hash or create a method that returns the type on the object

i.e.

``` ruby
    def store(*args)
      document, options = args
      type = get_type_from_document(document)

      if options
        percolate = options[:percolate]
        percolate = "*" if percolate === true
      end
```

In my opinion, type should always be allowed to be specified just like how you can when you remove from the index.
Personally, looking at the code, it seems kind of hacky the way the getting the type works. And furthermore, the way it is implemented means that you cannot pass in a json string to store, and specify the type as well. In fact a few weeks ago, a lot of my tests started to break because of this. Because there is no explicit way to set the type in store, I had a method that was returning the type for the index, but all of a sudden it started using "document" as the type, and I had no idea why. Looking at the code, it seems as though if you have a type that is the same as the type of class it will not use that type, why? Anyway, I had to (what feels to me like a hack) implement another method document_type that returned the type. But now I am wondering how long that is going to be stable as the type. And having this sort of way that you can specify the type just seems like not wanting to commit to something (for whatever reason) and just have a method signature like

``` ruby
store(type, document, *args)
```

That would have been the logical choice of method signature for me.

Let me know what you think.

karmi:
Hi, it took me a while to decipher what exact issue you are experiencing. I think I understand now and I am grateful for the issue, because it shed some light on important matters.

1/ You're right that the interface is asymmetric and absolutely _not_ elegant the way it is.

2/ I don't know what are you referring to as "hacky", though. Do you consider getting type from the document in [`Index#get_type_from_document`](https://github.com/karmi/tire/blob/master/lib/tire/index.rb#L265-279) as "hacky"? I tend to think the logic is very clear and the code is quite robust.

2/ The real problem is that `remove` is symmetric with `store`, and not with `retrieve`. In ElasticSearch, _get_ and _delete_ takes document type and document ID as arguments. This is something we'll fix as soon as possible.

[...]

PR #279, opened 2012-03-17
Add support for ElasticSearch's Update API, new in 0.19

alicial:
Using the Update API you don't have re-index the entire document, you can just set the fields that have changed with the new value, and the re-indexing is done on the server.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -240,6 +240,15 @@ module Tire
       logged('_percolate', curl)
     end
 
+    def update(type, id, options={})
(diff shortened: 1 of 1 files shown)


karmi:
Hi, thanks! Couple of notes:
- Could you please add unit and integration tests for the feature?
- I haven't decided yet how to work with `type` and `id` being passed to `Index` methods -- see #275. Anybody has a strong opinion on that?
- What about model integration? We probably want to to use the _Update API_ in `MyModel.tire.update_index`?
  * If we do that, how will it affect percolation and possibly other _Tire_ features?

PR #291, opened 2012-03-25
Fixes #290 bulk store escape broken

### LINKED ISSUE #290: URL escaped _type (for namespaced models) cannot be found
This is a problem with #218 and its associated changes
### The main problem here is that _type is being URL escaped when going into ES:

```
$ rake environment tire:import CLASS='Market::Property' FORCE=true
[IMPORT] Creating index 'development-realestate::application-properties' with mapping:
{"market/property":{"properties":{"id":{"type":"integer"} ... }}} 
# */!\ *This is not actually true -- _type != "market/property" -- instead ES saves _type as market%2Fproperty (see below) */!\*
```
### When trying to do a search, we get 0 results when we shouldn't be:

```
Market::Property.search({ ... })
curl -X GET "http://localhost:9200/development-realestate::application-properties/market%2Fproperty/_search?pretty=true" -d '{"query": {"match_all":{}}}'
# */!\*  You'll notice that "market%2Fproperty" is being used for _type, which obviously represents "market/property" */!\*
{
  "took" : 0,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "failed" : 0
  },
  "hits" : {
    "total" : 0,
    "max_score" : null,
    "hits" : [ ]
  }
}
```
### Let's take a look at one of the records

```
$ curl -X GET "http://localhost:9200/development-realestate::application-properties/_search?pretty=true" -d '{"query": {"match_all":{}}}'
# */!\* Just doing _search on all types for this index to retrieve records */!\*
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 5,
    "successful" : 5,
    "failed" : 0
  },
  "hits" : {
    "total" : 1000,
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "development-realestate::application-properties",
        "_type" : "market%2Fproperty", # >>>> */!\* the type is saved as a URL escaped string */!\* <<<<<
        "_id" : "100629",
        "_score" : 1.0, "_source" : { ... }
      }
    ]
  }
```
### Conclusion:

Since we are defining a URL escaped _type, we need to access it by double URL escaping (to translate the % sign). This is obviously pretty hacky & we should consider an alternative.

```
### Double escaped:
curl -X GET "http://localhost:9200/development-realestate::application-properties/market%252Fproperty/_search?pretty=true" -d '{"query": {"match_all":{}}}'
```

sarmiena:


karmi:
@sarmiena Cleaned up the commit, and merged. Thanks for looking into it, I think everything works fine with namespaced models now.

Also, I reverted changes to the ActiveRecord tests (not really helpful, seems to me? let's prefer simplicity in tests), and added a more lightweight unit test implementation.

PR #292, opened 2012-03-27
Keep given search payload in the body.

dylanahsmith:
Previously the payload would get encoded in the params as well as used in the body for a search. The payload would also be removed during the request, causing it to missing when the request is logged or if it is re-performed.

Regression test included.


dylanahsmith:
@karmi, any feedback?

saulius:
Please merge this in it's really annoying. Especially if one uses tire-contrib which logs the query and deletes payload from options making your real query a nonsense.

karmi:
It's again a “piecemeal” approach. The way _Tire_ works with options right now is ridiculous. I'd like to kill these problems in on sweep, have to find some time to review the codebase and find the sore spots.

erickt:
I seem to have run into the same problem, and I have a similar fix in #376. Adding a link here so I can review this commit later.

erickt:
Oops, forgot to link to my version of this [fix](https://github.com/erickt/tire/commit/4d2140541419f9bcf60a6c81268e7ce50a618bba).

PR #300, opened 2012-04-03
Added Dis Max Query

mszyndel:
As my company really needs dis max I added it.
Please let me know if I should write some more tests; units should be sufficient, but there are just two basic integration tests.


erickt:
Hi hajder! I think tire HEAD recently grew support for aliases. Any chance you could rebase onto head and update your patch?

mszyndel:
@erickt done.

mszyndel:
Yes, order of results is guaranteed.
I guess it would be nice if you read code before commenting on pull requests.

karmi:
@cwalsh You mean in tests? Yeah, the order should absolutely be guaranteed.

@hajder Don't let your bitterness get in front of polite discussion :) @cwalsh was just asking a question .)

test/integration/dis_max_queries_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+module Tire
+
+  class DisMaxQueriesIntegrationTest < Test::Unit::TestCase
+    include Test::Integration
+
+    context "Boolean queries" do
```
karmi: Should be `Dis Max queries`.

karmi:
I guess an integration test in this case would be great:
- It would clear any confusion (eg. about the guaranteed ordering, see discussion above)
- Integration tests serve as a great documentation on the exact purpose of the feature, not only as a "defensive mechanism". See eg. https://github.com/karmi/tire/blob/master/test/integration/index_update_document_test.rb#L25-35 [karmi/tire@ff00015] for an example for the _Update API_ which covers the usual use-cases for the API.

karmi:
Additional info for the integration test -- there's a great example of "albino elephant" at the http://www.elasticsearch.org/guide/reference/query-dsl/dis-max-query.html page, which should be easy to translate to a _Tire_ integration test:

> If the query is “albino elephant” this ensures that “albino” matching one field and “elephant” matching another gets a higher score than “albino” matching both fields.

mszyndel:
I wrote an additional test to check order of results returned, however it's not an "albino elephant" example (I belive this would be too dependent on elasticsearch internals).

@karmi I'm not completely sure if I did rebase right, so please take a look :)

karmi:
@hajder Thanks, merged.

@hajder, @nbudin Guys, maybe you can shed some light on your `dis_max` usage for me? As you can see from note in karmi/tire@7431ce9, the `dis_max` query gives the same result as a plain old `boolean` query? Can you share some example where it does make a difference from your application/domain? Thanks!

nbudin:
Interesting.  I hadn't been aware that a `boolean` query with all `should` clauses and `minimum_number_should_match: 1` would give the same result as a `dis_max` query; I had assumed the same thing you had at the bottom of that comment: that `boolean` would sum the scores.

I do think `dis_max` has the advantage of being both more specific and simpler.  It's also less likely to change in future ElasticSearch updates - not that `boolean` is necessarily likely to change, but `dis_max` should always mean dismax.

PR #308, opened 2012-04-09
Multi-Search support

romanbsd:
Hi,

I added the msearch support some time ago.
Unfortunately it seems that I will never have the time to add the specs (yes, I know, I should have done it before writing the code).
However it'll be pity if this is thrown away.


timuckun:
I have taken the approach of creating just one index and using bulk updating to put all items into the same index like this.

{ "id":1,"type:'this'},
{"id":1, "type":"that"}

Etc.  I guess that's not so great if you want to make specific types of indexes but I am wondering if there is something profoundly wrong with this approach.

romanbsd:
@timuckun there's nothing wrong with this approach, but it's unrelated to the multi search functionality. Multi search makes it possible to send different queries in one HTTP request. They can be to the same index or to different indeces.

timuckun:
Sorry I misunderstood.

karmi:
Another approach: http://spacevatican.org/2012/6/12/using-multisearch-with-tire

Gist: https://gist.github.com/92d77b62eb570031823d

timuckun:
Nice writeup, I'll probably implement this very soon. Thanks to whoever wrote it.

karmi:
@romanbsd Closing this, due to missing test suite, etc., all right? We'll need to add proper multi-search support...

karmi:
Closed in d233740, 775f537, c03c666

PR #314, opened 2012-04-14
Added Custom Filters Score Query support

jerryluk:
karmi,
Thanks for the wonderful library. Since we need to use Custom Filters Score Query for our project, we decided to go ahead and added the changes necessary for supporting Custom Filters Score Query. Here is how would an user to use it:

Model.tire.search(:load => true) do
  query do
    custom_filters_score do
      query { string query }
      filter do
        filter :term, :tags => "ruby"
        boost 2.0 # script "..." is also supported
      end
      score_mode "total"
    end
  end
end

We also added the test cases for it. We would like to contribute this back to the community, please let us know if there is any changes you would like us to make.

Thanks again.
- Jerry

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -210,7 +210,9 @@ module Tire
       end
       
       def to_hash
-        @value
+        @value[:filters].present? ? 
+        @value : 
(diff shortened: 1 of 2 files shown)


karmi:
Closed from karmi/tire-contrib@360348a

PR #317, opened 2012-04-19
Updated multijson deprecations introduced in MultiJson 1.3

johnpaulashenfelter:
I just went ahead and made the changes -- I had opened an issue about.

MultiJson 1.3 renamed some methods so I updated the dependency and did a search/replace. Tests passed before and after so looks like it's a simple fix. 

Not sure when/if you want to update the MultiJson dependency since its a big jump but figured other people would be bothered by the log messages as well.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -38,12 +38,12 @@ module Tire
 
     def mapping
       @response = Configuration.client.get("#{Configuration.url}/#{@name}/_mapping")
-      MultiJson.decode(@response.body)[@name]
+      MultiJson.load(@response.body)[@name]
     end
 
     def settings
       @response = Configuration.client.get("#{Configuration.url}/#{@name}/_settings")
-      MultiJson.decode(@response.body)[@name]['settings']
+      MultiJson.load(@response.body)[@name]['settings']
     end
 
     def store(*args)
@@ -62,7 +62,7 @@ module Tire
       url += "?percolate=#{percolate}" if percolate
 
       @response = Configuration.client.post url, document
-      MultiJson.decode(@response.body)
+      MultiJson.load(@response.body)
 
     ensure
       curl = %Q|curl -X POST "#{url}" -d '#{document}'|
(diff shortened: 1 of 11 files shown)


karmi:
I would prefer to wait on this issue a little bit longer. From the brief reading I did on the encode/decode vs. dump/load, it doesn't strike me as a particularly great example of responsible library design.

Directly changing all `encode` to `dump`, and forcing users on 1.3 doesn't strike me as an example of responsible library design on our part.

Rails does [couple of tricks](https://github.com/rails/rails/pull/5896/files) right now to detect which method to use etc. I agree that the deprecation warnings are obnoxious.

I think we should solve this when we're nearing the `0.5` release?

johnpaulashenfelter:
I tend to agree as far as the changes -- squashing the deprecations in the tests is my only real goal.

Do you think it's worthwhile to require MultiJson < 1.3? Not sure if there's anything in 1.3+ that makes the deprecations worth the upgrade -- you're probably in a better position for knowing that. That's what I did locally to get rid of the hassle Obviously that can be done on project by project basis in a gemfile was thinking _maybe_ it makes sense to match this version of tire to the multijson it expects.

Thoughts?

karmi:
I'm really conflicted. I don't like the reasoning given by _MultiJson_, that's for sure. See the discussion at https://github.com/intridea/multi_json/issues/36.

I'm not sure about the benefits of MultiJSON > 1.3, probably more encoders, such as `oj`... I tend to be very conservative with pushing updates on library users, there are tons of people out there using _Tire_ in various scenarios...

I think we could push for this in the `0.5` release -- in the end, it's really a matter of simple find and replace.

Or, we _could_ do what Rails does, laboriously detect which method to use, or use our _own_ wrapper around the _MultiJson_ wrapper -- but that would seem as the ultimate silliness :)

(I tried to set `test.verbose = false` in the _Rake_ tasks, but it does not get propagated further down the line to suppress `Kernel.warn`, it seems.)

johnpaulashenfelter:
I agree -- don't change req'd lib version until next release.

karmi:
So, let's close this and reopen when we're nearing 0.5 and the dust on load/dump settles?

johnpaulashenfelter:
Agreed

PR #323, opened 2012-04-25
Add debug output to connection errors

dennisreimann:
Makes it easier to debug connection errors by adding the url to raised ConnectionExceptions

### DIFF
--- a/lib/tire/http/client.rb
+++ b/lib/tire/http/client.rb
@@ -5,52 +5,37 @@ module Tire
     module Client
 
       class RestClient
-        ConnectionExceptions = [::RestClient::ServerBrokeConnection, ::RestClient::RequestTimeout]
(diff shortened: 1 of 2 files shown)


karmi:
Yes, this makes sense, thanks for that!

As I said in many previous cases, these kinds of patches are half-solutions, before we use some better HTTP abstraction, notably _Faraday_. Notice that your patch is only for the _RestClient_ client (not what anybody should use in production), and that we're nearing our custom HTTP abstraction. That's not a place I'd like to go...

I will make some time for refactoring the HTTP support in the coming months. I'll meditate over this patch a bit in the mean time :)

dennisreimann:
Thanks for your feedback. While working on this patch I noticed that a proper HTTP abstraction would be the right way to go, you are absolutely right. In the meantime, this already helps us to debug connection issues we are facing sometimes.

nz:
This might be a half solution, but it looks pretty un-invasive and very helpful in the context of issues like #324.

:+1:

PR #328, opened 2012-04-30
Index should bulk store to its own URL endpoint (fixes #327)

### LINKED ISSUE #327: Bulk import with Bonsai.io/Heroku and Tire
I have setup Bonsai on my Heroku app following this gist: https://gist.github.com/2041121, and everything seems to work except for the bulk import of existing posts.  Any new post gets put into the ES index automatically, but when I try to import old posts, either directly from the Heroku console, e.g.:

``` ruby
posts = Post.limit(100)
Tire.index BONSAI_INDEX_NAME do
  import posts
end
```

 or by running `heroku run rake environment tire:import CLASS=Post FORCE=true`, I get this error:

`[ERROR] Too many exceptions occured, giving up. The HTTP response was: 401 > {"error": "Not authorized: Some endpoints are admin-only, ask support@onemorecloud.com."}`

One strange feature is that _some_ of my old posts get imported, but not all, and I can't see any reason why some fail while others succeed.  Has anyone else had any luck doing a bulk import into Bonsai?  Many thanks for any help!

cc @nz

nz:
The `Index#bulk_import` method  ([index.rb:123](https://github.com/karmi/tire/blob/master/lib/tire/index.rb#L123)), and other index methods, interact with the Cluster URL rather than their own Index URL endpoint.

This pull request refactors the index URL endpoint into the `Index#url` getter method, refactors a number of methods to use this new getter for clarity, and changes the `bulk_import` method to use this index-level endpoint rather than the cluster-level URL endpoint.

Also updated/refactored test expectations accordingly.

This fixes issues (e.g., #327) for users trying to configure Tire to use a single application index and who do not have access to cluster-level methods like `/_bulk` in production.

---

If anyone following along wants to test, try this in your `Gemfile`:

``` ruby
gem 'tire', git: 'git@github.com:nz/tire', branch: 'index-bulk-import-to-index-endpoint'
```

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -8,31 +8,35 @@ module Tire
       instance_eval(&block) if block_given?
     end
 
+    def url
+      "#{Configuration.url}/#{@name}"
+    end
+
     def exists?
-      @response = Configuration.client.head("#{Configuration.url}/#{@name}")
+      @response = Configuration.client.head(url)
       @response.success?
 
     ensure
-      curl = %Q|curl -I "#{Configuration.url}/#{@name}"|
+      curl = %Q|curl -I "#{url}"|
       logged('HEAD', curl)
     end
 
     def delete
-      @response = Configuration.client.delete "#{Configuration.url}/#{@name}"
+      @response = Configuration.client.delete(url)
       @response.success?
 
     ensure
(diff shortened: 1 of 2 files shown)


nz:
thanks! :heart:

PR #332, opened 2012-05-04
Add support for update script API on Tire::Index.

ahfeel:
Add support for update script API on Tire::Index.
Usage:
tags = ['a', 'b', 'c']
Tire.index 'my_index' do
  update 'document_type', 'document_id', {
    :script => "_ctx._source.tags = tags;',
    :params => {:tags => tags}
  }
end

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -296,6 +296,19 @@ module Tire
       logged('_percolate', curl)
     end
 
+    def update(type, id, options)
+      raise ArgumentError, "Please pass a document ID" unless id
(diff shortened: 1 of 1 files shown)


karmi:
Hi, thanks. Couple of remarks:

1/ `Index#store` uses args introspection to get the type and id. It should change to allow more flexibility and I have to think about that.

2/ Could you add at least a unit test for the feature. I can write integration test myself.

3/ It's more convenient when you _rebase_ against current master in Git workflows. I can get rid of the merge and rebase myself, but still, better when it's not there.

karmi:
@ahfeel Any news?

PR #334, opened 2012-05-07
Recent ES doesn't like the sort parameter, so it seems.

romanbsd:
When the query is provided as hash, the params get concatenated, producing a query like this:

```
curl -X GET "http://localhost:9200/graylog2/message/_search?from=0&page=1&per_page=100&size=100&sort=created_at+desc&pretty=true" -d '{"query":{"query_string":{"query":"*"}},"sort":[{"created_at":"desc"}],"size":100,"from":0}'
```

It produces an error:

Parse Failure [No mapping found for [created_at desc] in order to sort on]

I don't know whether the other parameters should be there, as they're duplicated in the query, but at least they don't make the request invalid.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -72,8 +72,8 @@ module Tire
             options ||= {}
           end
 
-          sort      = Array( options[:order] || options[:sort] )
           options   = default_options.update(options)
+          sort      = Array( options.delete(:order) || options.delete(:sort) )
 
           s = Tire::Search::Search.new(options.delete(:index), options)
           s.size( options[:per_page].to_i ) if options[:per_page]
@@ -298,7 +298,6 @@ module Tire
         Results::Item.send :include, Loader
       end
 
-      
     end
 
   end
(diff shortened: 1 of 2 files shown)


karmi:
Hi Roman, yeah, that's a bug -- I'd rather have an overall look over the codebase to rip instances like these out..

romanbsd:
Maybe you can merge this meanwhile, as otherwise the 0.4.x tire is unusable with graylog2-web-interface. Unless, of course, you are going over the codebase in the near future ;)

karmi:
Is it possible to add a simple unit test covering the bug?

romanbsd:
Added a test.

karmi:
@karmi @vhyza Bump.

romanbsd:
bump

elfassy:
Is sorting working now?

odino:
+1, graylog web interface totally not working

karmi:
Merged it in, moved the test to a better place, it should not break anything. Sorry it took so long to have a look at this...

PR #337, opened 2012-05-07
Faraday adapter

romanbsd:
Using Tire with Faraday allows the swap the HTTP client with some other implementation. In our case, we're going to use http-net-persistent.

### DIFF
--- a/lib/tire/http/clients/faraday.rb
+++ b/lib/tire/http/clients/faraday.rb
@@ -1,5 +1,18 @@
 require 'faraday'
 
+# Example:
+#
+# require 'tire/http/clients/faraday'
+# Tire.configure do |config|
+#   # Unless specified, tire will use Faraday.default_adapter and no middleware
+#   Tire::HTTP::Client::Faraday.faraday_middleware = Proc.new do |builder|
+#     builder.adapter :net_http_persistent
+#     builder.use FaradayMiddleware::Instrumentation, :name => 'request.tire'
+#   end
+#
+#   config.client(Tire::HTTP::Client::Faraday)
+# end
+
 module Tire
   module HTTP
     module Client
--- a/lib/tire/http/clients/faraday.rb
+++ b/lib/tire/http/clients/faraday.rb
@@ -1,0 +1,47 @@
+require 'faraday'
+
+module Tire
+  module HTTP
+    module Client
+      class Faraday
+
+        # Default middleware stack.
+        DEFAULT_MIDDLEWARE = Proc.new do |builder|
+          builder.adapter ::Faraday.default_adapter
+        end
+
+        class << self
+          # A customized stack of Faraday middleware that will be used to make each request.
+          attr_accessor :faraday_middleware
+
+          def get(url, data = nil)
+            request(:get, url, data)
+          end
+
+          def post(url, data)
+            request(:post, url, data)
+          end
+
+          def put(url, data)
+            request(:put, url, data)
+          end
+
+          def delete(url, data = nil)
+            request(:delete, url, data)
+          end
+
+          def head(url)
+            request(:head, url)
+          end
+
+          private
+          def request(method, url, data = nil)
+            conn = ::Faraday.new( &(faraday_middleware || DEFAULT_MIDDLEWARE) )
+            response = conn.run_request(method, url, data, {})
+            Response.new(response.body, response.status, response.headers)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/tire/http/clients/faraday.rb
+++ b/lib/tire/http/clients/faraday.rb
@@ -37,7 +37,7 @@ module Tire
           private
           def request(method, url, data = nil)
             conn = ::Faraday.new( &(faraday_middleware || DEFAULT_MIDDLEWARE) )
-            response = conn.run_request(method, url, data, {})
+            response = conn.run_request(method, url, data, nil)
             Response.new(response.body, response.status, response.headers)
           end
         end


karmi:
Hey, nice idea to add _Faraday_ as one of the regular “clients” :) As I've mentioned in couple of places, I'd like to switch to _Faraday_ as the HTTP abstraction very soon.

I'm not against merging this in the meantime -- I'd just like some more detailed information, preferably in the `faraday.rb` file itself, how to set up the adapter, etc etc. Tests are not neccessary in this case.

romanbsd:
I added a configuration example

karmi:
Hi Roman, looks good, will have a look at that shortly.

ghost:
+1 to subscribe

ghost:
@romanbsd I just deployed this and found that all my simple queries return the first X records in the index no matter what I searched for. Returning back to the Restclient client, restores functionality. I don't see anything obviously different about the #get request, but I'm much less familiar with why that might be. 

Did you see any of this?

romanbsd:
No, I didn't experience something like this. On the other hand, our tire has some more modifications (though they are not related).
Can you sniff the trafic with tcpdump or similar?
P.S. We're using faraday 0.8.0

ghost:
The working client shows this for my search:

00000000  7b 22 71 75 65 72 79 22 3a 7b 22 71 75 65 72 79 5f 73 74 72 69 6e 67 22 3a 7b 22  {"query":{"query_string":{"
0000001b  71 75 65 72 79 22 3a 22 67 6f 6f 67 6c 65 2a 22 7d 7d 2c 22 73 6f 72 74 22 3a 5b  query":"google*"}},"sort":[
00000036  7b 22 63 6f 75 6e 74 22 3a 22 64 65 73 63 22 7d 5d 2c 22 66 69 6c 74 65 72 22 3a  {"count":"desc"}],"filter":
00000051  7b 22 72 61 6e 67 65 22 3a 7b 22 63 6f 75 6e 74 22 3a 7b 22 67 74 22 3a 32 7d 7d  {"range":{"count":{"gt":2}}
0000006c  7d 2c 22 66 69 65 6c 64 73 22 3a 5b 22 6e 6f 72 6d 61 6c 69 7a 65 64 5f 6e 61 6d  },"fields":["normalized_nam
00000087  65 22 5d 7d                                                                       e"]}    

This branch, show nothing in the same place.

ghost:
I looked into Faraday and it appears that if you create a connection object, the `run_requeset` method is triggered here: https://github.com/technoweenie/faraday/blob/master/lib/faraday/connection.rb#L159

For `get` requests, it blanks out the body, assuming it doesn't need the data, which is exactly the behavior I'm seeing.

I'm not sure how this would actually work for you without changing Faraday. Do you get body content on a `get` request?

erickt:
@brandonmeeteor: Changing Tire to use POSTs instead of GETs for requests with bodies should work.

romanbsd:
@brandonmeeteor I'm calling the run_request itself, https://github.com/technoweenie/faraday/blob/master/lib/faraday/connection.rb#L386 , so the body should be present.

romanbsd:
I verified here that the following works as expected:

``` ruby
require 'faraday'
c = Faraday.new
c.run_request(:get, 'http://localhost:3333/', '{foo:"foo"}', {})
```

I also tried:

``` ruby
Tire.search {|s| s.sort { by :date }; s.query {|q| q.term 'bar', 'baz'} }.perform
```

and I'm seeing:

<pre>
GET /_search HTTP/1.1
Accept: */*
User-Agent: Ruby
Connection: close
Host: localhost:9200
Content-Length: 57
Content-Type: application/x-www-form-urlencoded

{"query":{"term":{"bar":{"term":"baz"}}},"sort":["date"]}
</pre>


That makes me think, that the problem is with the faraday adapter. What Faraday adapter are you using? We're using net_http_persistent. Other adapters, such as typhoeus or patron might not send the body.

ghost:
Ah...gotcha. Looks like our gemset was relying on Faraday 0.7.5 from the Koala gem. I removed it and update Faraday to the latest, and the query worked as expected.  Sorry for the mixup. I guess the adapter would have to be dependent on the latest though. Probably a valuable piece of info to know.

We can't upgrade the other gems that rely on Faraday quite yet, so I wrote this instead (https://github.com/karmi/tire/issues/351). It should work for the time being. I'll keep an eye on this one and upgrade everything else once this is merged in.

Thanks!

dpaluy:
@romanbsd +1 Good example.

karmi:
@romanbsd Thanks, Roman, merged!

lib/tire/http/clients/faraday.rb
```diff
@@ -0,0 +1,60 @@
+require 'faraday'
+
+# Example:
+#
+# require 'tire/http/clients/faraday'
+# Tire.configure do |config|
+#   # Unless specified, tire will use Faraday.default_adapter and no middleware
+#   Tire::HTTP::Client::Faraday.faraday_middleware = Proc.new do |builder|
+#     builder.adapter :net_http_persistent
+#     builder.use FaradayMiddleware::Instrumentation, :name => 'request.tire'
```
karmi, line 10: @romanbsd Roman, I'm wondering about this line. There's no match for `Instrumentation` nor `FaradayMiddleware` in `faraday` 0.8.1. I had to remove it to make it work. What's going on here?

romanbsd, line 10: This is just an example. The FaradayMiddleware is a separate beast. I just copied our production setup verbatim as an example of what's possible to do with this adapter.
https://github.com/pengwynn/faraday_middleware

karmi:
OK, changed the usage instructions to use `typhoeus` and removed the midleware part.

PR #342, opened 2012-05-12
Enhanced paginate method to sort by id by default

### LINKED ISSUE #272: Import broken
Tire::Index#import iteration algorithm does not guarantee unique records. Using kaminari & Postgres, I get SQL commands like this:

`SELECT  "products".* FROM "products"  LIMIT 1000 OFFSET 1000`

Note that without an `order by` clause, OFFSET will return random records since it doesn't have a concept of a cursor. Even kaminari addresses this concern in their documentation: https://github.com/amatsuda/kaminari/wiki/Kaminari-recipes

Also, why the need to use an external gem (will_paginate or kaminari) to iterate over records when ActiveRecord already has #find_in_batches and Mongoid has #batch_size (https://github.com/mongoid/mongoid/issues/1334)

kalarani:
This makes sure that there is a default ordering so that 'paginate' method doesn't fetch duplicate records for indexing. This commit fixes #272

### DIFF
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -73,7 +73,7 @@ namespace :tire do
       if defined?(Kaminari) && klass.respond_to?(:page)
         klass.instance_eval do
           def paginate(options = {})
-            page(options[:page]).per(options[:per_page]).to_a
+            order("#{klass.table_name}.id").page(options[:page]).per(options[:per_page]).to_a
           end
         end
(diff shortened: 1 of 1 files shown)


karmi:
I suspect this to be ActiveRecord specific? We need to support generic ActiveModels OxMs... Will close this unless you have a generic solution, OK?

karmi:
Closing, ActiveRecord specific code.

PR #354, opened 2012-05-29
This fixes an issue where klass.find does not return an array.

hdm:
Details of the problem can be found at http://stackoverflow.com/questions/10592005/tire-gem-undefined-method-detect


karmi:
Hey, I'm not able to reproduce the issue you describe with a test such as:

``` ruby
        should "return collection for one matching result" do
          results = ActiveRecordArticle.search :load => true do
              query { string '"Test 1"' }
            end

          assert_equal 1, results.size
        end
```

Can you provide a failing test or a more informative description of your data set etc?

karmi:
BTW, was tempted to fix it with `klass.find(Array(ids))`, but really can't reproduce it...

karmi:
Can't reproduce, closing...

PR #355, opened 2012-05-30
Added attr_accessible to Article model in rails-application-template

ghost:
Latest rails version requires attr_accessible to be able to update / insert given model.

Added attr_accessible to Article model.


karmi:
Thanks!, merged.

Issue #244, opened 2012-02-16
Implement script_fields


mattiassvedhem:
Would it be possible to implement script_fields? If so, where would they best be implemented?

[Script fields](http://www.elasticsearch.org/guide/reference/api/search/script-fields.html)

I'm looking to mix something like this in mostly all my queries:

```
curl -XGET 'http://127.0.0.1:9200/geonames/_search?pretty=1'  -d '
{
   "script_fields" : {
      "distance" : {
         "params" : {
            "lat" : 2.27,
            "lon" : 50.3
         },
         "script" : "doc[location].distanceInKm(lat,lon)"
      }
   }
}
'

# [Thu Feb 16 11:20:29 2012] Response:
# {
#    "hits" : {
#       "hits" : [
#          {
#             "_score" : 1,
#             "fields" : {
#                "distance" : 466.844095463887
#             },
#             "_index" : "geonames_1318324623",
#             "_id" : "6436641_en",
#             "_type" : "place"
#          },
```

karmi:
Yeah, I guess it's on the same level as the `field` method, https://github.com/karmi/tire/blob/master/lib/tire/search.rb#L75-78

PR #363, opened 2012-06-03
Script fields

martinciu:
Hi,

It adds support for script_fields (http://www.elasticsearch.org/guide/reference/api/search/script-fields.html). 
#244


karmi:
@martinciu Excellent patch, thanks!!

PR #367, opened 2012-06-04
More index store options

fcheung:
Support all the option you can pass when indexing individually or in bulk.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -85,15 +85,16 @@ module Tire
     end
 
     def bulk_store(documents, options={})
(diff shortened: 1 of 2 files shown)


jerryluk:
Would Love to have parent support. +1

mattiassvedhem:
+1.

fcheung:
Yeah bulk_store is a little messy. It's got a lot in common with multi_search if my memory is correct so there is probably an opportunity to find a nice solution to deal with both.

karmi:
Closing in favour of more complete implementation...

Issue #364, opened 2012-06-03
Merge stored attributes with loaded model attributes


martinciu:
Hi,

I'd like to be able to merge stored attributes ( from `_source` or `fields`) into loaded model (via `:load => true` flag). Is there easy way to do it? I was thinking of wrapping loaded model into some `ResultDecorator`. 

What do you think about it?

Thanks!

karmi:
Duplicate of #368

PR #261, opened 2012-03-05
add possibility to bind the raw results to the eager loaded model

phoet:

PR #368, opened 2012-06-05
Result decorator

martinciu:
Hi,

As I described in #364 I needed a way to merge attributes returned by ElasticSearch with model loaded from database. I needed to get merge them with values returned from `scritp_fields` (see #363). 

So now search results returns array of `Results::Item` whenever you get load models for results or not and passes missing methods first to `attributes` array and than to loaded model.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -1,4 +1,24 @@
 module Tire
+  class UnknownModel < Exception
+    def initialize(type)
+      @type = type
+    end
+
+    def message
+      "You have tried to eager load the model instances, but " +
+      "Tire cannot find the model class '#{@type.camelize}' " +
+      "based on _type '#{@type}'."
+    end
+  end
+
+  class UnknownType < Exception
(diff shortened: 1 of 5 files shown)


karmi:
@martinciu Hi, I like the refactoring you did, sorry it took me so long to get to it.

I'm not sure about returning `Item` instances when using the `:load` option, though. I think we should have something like Sunspot's `each_with_hit` (see karmi/tire#406 and karmi/tire#261).

I'll try to have a look into this shortly.

martinciu:
Hi @karmi I'll take a look at Sunspot's `each_with_hit`

thanks!

karmi:
@martinciu What do you think about the `each_with_hit` approach? I think it would be better then to decorate the results directly? Also notice the refactoring I did in https://github.com/karmi/tire/tree/refactoring/results to better support people needing to change the code which loads data from the database...

karmi:
Closing in favour of `each_with_hit` (or similar) feature.

PR #374, opened 2012-06-11
Mapping generated automatically shouldn't have type property

andhapp:
If we use the following:

<pre>
Tire.index 'articles' do
  store :type => 'article',
          :title => 'Three',
          :tags => ['java'],
          :published_on => '2011-01-02'
end
</pre>

then articles index has 'type' as one of the properties in the mapping, for example: 

<pre>
{"analytics":{"study_tracking":{"properties":{"id":{"type":"string"},"member_id":{"type":"string"},"tracked_on":{"type":"date","format":"dateOptionalTime"},"type":{"type":"string"}}}}}
</pre>

This pull request fixes that so that type is not one of the properties in the mapping.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -305,7 +305,7 @@ module Tire
         when document.respond_to?(:document_type)
           document.document_type
         when document.is_a?(Hash)
-          document[:_type] || document['_type'] || document[:type] || document['type']
+          document.delete(:_type) || document.delete('_type') || document.delete(:type) || document.delete('type')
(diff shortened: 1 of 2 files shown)


karmi:
There's still lots of issues around storing or not `_type` in the `_source`, as far as I can see this does not create any problems: it's just _another_ property in source, that's all. This issue must be dealt with in a much more thorough way.

Closing for now.

andhapp:
I can't remember the issue we had related to this as it's been a while. 

Thanks, anyways.

Issue #205, opened 2012-01-15
Boosting Entire Indices using Tire


rozumem:
Does Tire provide an wrapper method to boost entire indices? 

For example, if I have three indices and want to search across all of them, I want certain indices to be given greater preference. http://www.elasticsearch.org/guide/reference/api/search/index-boost.html

Couldn't find it in the Tire documentation, does a wrapper exist?

Thanks.

vhyza:
Hello,

I'm afraid indices_boost option is not supported yet.

karmi:
Maybe something like this would make sense from the _Tire_'s DSL point of view:

``` ruby
Tire.search 'index_1' => { boost: 1.5 }, 'index_2' => { boost: 2 }
```

andruby:
I want this feature too. I'm going to look around in the Tire code to see if I can pull it off.

PR #377, opened 2012-06-16
Boost entire indices

andruby:
Added the ability to boost entire indices.
http://www.elasticsearch.org/guide/reference/api/search/index-boost.html

Using the DSL proposed by karmi in #205.

``` ruby
Tire.search 'index_1' => { boost: 1.5 }, 'index_2' => { boost: 2 }
```

PS @karmi: I love how clean the code is and how easy it was to add this feature.


karmi:
Thanks, merged!

And thanks for the kind words, I don't see the codebase in such a positive light at all :), but am glad you enjoyed working with it!

PR #379, opened 2012-06-16
Scan now accepts :scroll option. 

mrazicz:
Scan fixed to accepts :scroll option. Default is '10m'.


karmi:
Hi, the scan feature will be refactored due to scan/scroll mixup, see https://github.com/erickt/tire/commit/16ca8ba1f2949b2d4033c021e9477f350d8799b6

mrazicz:
Thx for the quick reply. So i think this pull request is useless and i close it.

PR #383, opened 2012-06-19
Adds support for classes that don't respond to :new, e.g. Integer

cwalsh:
For example, with the below class, the Integer conversion now works. Before it would try to call `Integer.new('4')`, now it tries to call `Integer('4')`.

``` ruby
class PersistentArticleWithCastedNumber

  property :title
  property :count, :class => Integer
  property :stats
end
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -11,7 +11,7 @@ communicating by JSON over RESTful HTTP, based on [Lucene](http://lucene.apache.
 
 This Readme provides a brief overview of _Tire's_ features. The more detailed documentation is at <http://karmi.github.com/tire/>.
 
-Both of these documents contain a lot of information. Please set aside some time to read them thoroughly, before you blindly dive into „somehow making it work“. Just skimming through it **won't work** for you. For more information, please refer to the [integration test suite](https://github.com/karmi/tire/tree/master/test/integration)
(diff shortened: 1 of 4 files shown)


romanbsd:
What about Fixnum?

erickt:
For reference, I have a similar patch in my pull request, but I take the approach of hardcoding the types to cast:

https://github.com/erickt/tire/commit/2281af5f836e4cbf5bfbed823837fec28281e9b2

Which do you all like better?

cwalsh:
Romanbsd: Fixnum is an interesting one, see http://www.rubyfleebie.com/understanding-fixnums/ for reasons.  However, if you use Integer, it gets converted to Fixnum anyway.

Erickt, my patch doesn't handle Time, I might look in to it later. Does yours work with DateTime etc.?

cwalsh:
Ok, https://github.com/karmi/tire/pull/352/files#diff-5 got merged so this is now irrelevant.

karmi:
@cwalsh That wasn't yet merged, please allow me couple of days to process the pipeline :)

BTW, you have changed the commit, right? Now the tests are much more lucid about the intent.

cwalsh:
Oh whoops, I misread that and thought #352 was merged already, but it was just closed.  Yes, I've updated the pull request because I initially missed the option for :class => [Float] not just :class => Float.  Thanks for reopening.

PR #388, opened 2012-06-21
Support for span_near / span_or / span_term queries

squidfunk:
Hi there,

I was in need of support for span_near / span_or / span_term queries, so I added it:

```
MyIndex.search do
  query do
    span_near slop: 10, in_order: true do
      span_term :field1, value1
      span_term :field2, value2
    end
  end
end
```

If you find it useful, you're free to merge it.

All the best,
Martin

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -42,6 +42,13 @@ module Tire
         end
       end
 
+      def custom_boost_factor(options={}, &block)
+        @custom_boost_factor ||= Query.new(&block)
+        @value[:custom_boost_factor] = options
+        @value[:custom_boost_factor].update({:query => @custom_boost_factor.to_hash})
+        @value
+      end
+
       def custom_score(options={}, &block)
         @custom_score ||= Query.new(&block)
         @value[:custom_score] = options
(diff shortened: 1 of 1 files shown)


karmi:
Hi, this is really rather heavy for the "every query in one file"... care to submit it on `tire-contrib`? I guess you have it loaded just as an extension for Tire, right?

squidfunk:
Hey, I don't really know how to write extensions for tire - is there a guide somewhere? As you see, I put it into the actual `query.rb`. However, I could contribute it to tire-contrib when I find the time.

karmi:
It's really easy -- have a look on the [existing queries](https://github.com/karmi/tire-contrib/blob/master/lib/tire/queries/custom_filters_score.rb), it's just a matter of extending the Tire classes, very similar to just "pasting" it inside query.rb.

(Which you shouldn't do, BTW :) It would work equally well if you would have put it into a new file, and just `require`-d that after Tire...)

karmi:
Closing due to pending pull request on Tire Contrib.

squidfunk:
Sounds absolutely legit! When I find the time (fighting on other frontiers at the moment), I will contribute this to tire-contrib if no one has already done this. Thanks for your feedback!

karmi:
Thanks!

PR #403, opened 2012-07-10
Use standard Logger interface in Tire::Logger

1st8:
Hi,

when developing in a Virtualbox shared folder, the `Tire::Logger` sometimes fails to write to the logfile with `Errno::EPROTO` (in #write).

I made a quick patch that replaces the device with a ruby Logger instance, which works fine on shared folders.

Regards

1st8

### DIFF
--- a/lib/tire/logger.rb
+++ b/lib/tire/logger.rb
@@ -1,15 +1,13 @@
 module Tire
   class Logger
 
-    def initialize(device, options={})
-      @device = if device.respond_to?(:write)
-        device
+    def initialize(logger, options={})
+      @logger = if logger.is_a?(::Logger)
+        logger
       else
-        File.open(device, 'a')
+        ::Logger.new(logger)
       end
-      @device.sync = true if @device.respond_to?(:sync)
       @options = options
-      # at_exit { @device.close unless @device.closed? } if @device.respond_to?(:closed?) && @device.respond_to?(:close)
     end
 
     def level
@@ -17,7 +15,7 @@ module Tire
     end
 
     def write(message)
-      @device.write message
+      @logger.send(level.to_sym, message)
     end
 
     def log_request(endpoint, params=nil, curl='')


karmi:
Hi, thanks! Logger injection is something of a sore spot still.

Notice this works:

``` ruby
require 'logger'
require 'tire'

class Logger
  alias :write :<<
  def sync=(*args)
  end
  def closed?
  end
end

Tire.configure do
  logger Logger.new('ruby-logger-tire.log', 3, 1024)
end

Tire.search do
  query { string '*' }
end
```

And also this:

``` ruby
require 'log4r'
require 'tire'

mylog = Log4r::Logger.new 'Tire'
mylog.outputters = Log4r::FileOutputter.new 'Tire', :filename => 'log4r-tire.log'
mylog.instance_eval do
  alias :write :info
end


Tire.configure do
  logger mylog
end

Tire.search do
  query { string '*' }
end
```

And also this:

``` ruby
require 'fileutils'
require 'active_support/buffered_logger'
require 'tire'

mylog = ActiveSupport::BufferedLogger.new 'buffered-logger.log'
mylog.instance_eval do
  alias :write :info
end

Tire.configure do
  logger mylog
end

Tire.search do
  query { string '*' }
end
```

Notice the silly monkeypatching with `write`, `closed`, etc.

I think we should leave out the logger injection/definition to the user? Do you think you can improve the logger support to be more generic in this sense?

Or should we just stick to the default `Logger` interface? Anybody has any strong opinion on this?

1st8:
Hi, thanks for your detailed response :)

What are the advantages of your solution against a default `Logger` interface implementation?
If there are none I would prefer the "standard" solution.

I saw a simple example in mongoids implementation https://github.com/mongoid/mongoid/blob/master/lib/mongoid/loggable.rb
For tires log output it could be better to omit the `Rails.logger` usage ;)

I can implement it later today or tomorrow if you want.

karmi:
@1st8 Sorry for the delay -- yeah, I think my original code is probably a monument of overthinking it and using just the `Logger` makes sense. I'll try to look into it shortly!

dirkbolte:
:+1: I just hit a similar issue with using logger in tire. In my setup, I have to use a Lumberjack logger. Getting this working with tire resulted in a similar implementation as shown above. Having a way to assign a custom logger which sticks to the Logger interface would be great. 
An alternative to the mongoid implementation would be: https://github.com/vigetlabs/loggable - it provides a stub per default which ignores logs, so instead of checking for null, you can actually provide a block which build the message - which would not be executed if no Log object is assigned.
I personally prefer being able to assign a logger. If there are concerns with breaking the current interface, I can try to provide an implementation which supports both approaches.

PR #407, opened 2012-07-17
Formatting

richo:
I fixed a typo (operation system) and nuked all the trailing whitespace in index.html


karmi:
Hi, good catch with the typo, but the `index.html` file is generated by [_Rocco_](http://github.com/rtomayko/rocco) from the  [`examples/tire-dsl.rb`](https://github.com/karmi/tire/blob/master/examples/tire-dsl.rb) file...

PR #415, opened 2012-07-25
dis_max support in DSL

nbudin:
This change implements support for the `dis_max` query type.  So you can do, for example:

``` ruby
Search::Search.new('index') do
  query do
    dis_max do
      query { string 'foo' }
      query { terms :tags, ['baz'] }
    end
  end
end
```

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -68,6 +68,13 @@ module Tire
         @value
       end
 
+      def dis_max(&block)
+        @dis_max = DisMaxQuery.new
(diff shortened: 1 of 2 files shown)


karmi:
Hi, thanks!, unfortunately an exact duplicate of karmi/tire#300...

nbudin:
Ah, ok!  I had searched for dis_max in the issue tracker to see if anyone else had done this, but apparently I should have searched for "dis max" instead. :)

No worries in any case, this didn't take me long to do and looks like the other pull request has had a lot more thought put into it than mine.

karmi:
Thanks, I'll make sure to link to your patch in the commit message at least...

PR #420, opened 2012-07-31
Fix for Tire::Alias interface for aliases including routing values

mriley:
This is a small change to the way Tire picks out the "indices" attribute passed to the Alias initializer. The original regex match only works if an existing Alias doesn't have a routing value associated with it. If a routing value is set, ElasticSearch returns a key called "index_routing", which can overwrite the "indices" or "index" value if it comes later in the hash iteration, since that key also matches the regex "/index|indices/". 

This problem appeared to us when using Tire::Alias.find() to find the indices associated with a particular alias. We were finding the alias correctly, but the list of indices returned with the Alias object was always incorrect — it was set to the value returned as "index_routing" instead.

Let me know if you need more details for reproduction.

### DIFF
--- a/lib/tire/alias.rb
+++ b/lib/tire/alias.rb
@@ -77,7 +77,7 @@ module Tire
       @attributes = { :indices => IndexCollection.new([]) }
 
       attributes.each_pair do |key, value|
-        if key.to_s =~ /index|indices/
+        if ['index','indices'].include? key.to_s
           @attributes[:indices] = IndexCollection.new(value)
         else
           @attributes[key.to_sym] = value
--- a/test/unit/index_alias_test.rb
+++ b/test/unit/index_alias_test.rb
@@ -83,7 +83,7 @@ module Tire
       context "updating" do
         setup do
           Configuration.client.expects(:get).
-                               returns( mock_response(                               %q|{"index_A":{"aliases":{}},"index_B":{"aliases":{"alias_john":{"filter":{"term":{"user":"john"}}},"alias_martha":{"filter":{"term":{"user":"martha"}}}}},"mongoid_articles":{"aliases":{}},"index_C":{"aliases":{"alias_martha":{"filter":{"term":{"user":"martha"}}}}}}|), 200 ).at_least_once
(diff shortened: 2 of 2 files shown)


karmi:
Hi, thanks, looks like there may be a bug -- could you please a unit test demonstrating the issue?

karmi:
@mriley Is it possible to add a [unit test](https://github.com/karmi/tire/blob/master/test/unit/index_alias_test.rb) for the fix? If you'd need some help with that, ping me.

mriley:
@karmi — Sorry for the delay. I'll try to knock out a test for it today and submit.

mriley:
I just made a few small changes to the mock responses in the Alias test suite that should cause errors with the existing version of Tire before the patch in my first commit is applied. The issue is that you can create an alias with a single "routing" value or with two separate routing values "index_routing" and "search_routing". Regardless of which you choose, the routing values are stored separately in ES as "index_routing" and "search_routing" (if you simply pass "routing" when you create it the same value is used for both routings). I added "index_routing" and "search_routing" values to the mock responses, and the "index_routing" value breaks the code I modified in my patch.

Fwiw, I think it makes sense for Tire to expose two different helper methods for setting the index_routing and search_routing values separately, but I thought that was more appropriate as a second patch. I will try to put that together along with some new tests.

karmi:
Bump @karmi.

mriley:
Did you have questions about the changes I made to the mock responses in tests? Those changes expose the original bug, and my patch fixes it.

karmi:
@mriley No no, this is just a reminder for me to have a look at it soon :)

mriley:
Oh ok :) Looking forward to seeing it in there soon. 

Let me know if you want to chat about other changes to the Alias interface. Like I said in my earlier comment, we might want to expose two different helper methods for setting the index_routing and search_routing values separately. I was having a hard time finding a way to make that change fully backward-compatible, though.

karmi:
@mriley Finally closed, sorry for the embarrassing delay with this. I've changed the test case, extracted it into a separate context -- it does not run _all_ the tests, but since the bug is in the initializer, it should catch it fine.

I'm not worried about backwards incompatible change if it makes sense. Setting up the index/search routings should be just:

```
Tire::Alias.new name: 'index_anne', ..., index_routing: '1', search_routing: '2'
```

Also, the `search_routing()` and `index_routing()`  would have to be added.

PR #421, opened 2012-08-01
Add ability to pass arguments to query { all }

woahdae:
In elasticsearch, match_all takes arguments that affect the boosting.
For example, if you want to perform empty queries that still respect
document boost, you need to pass {'norms_field' => '[field]'} to match_all.

This adds an optional hash argument to Search::Query#all that is a
pass-through to elasticsearch' match_all.

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -75,8 +75,8 @@ module Tire
         @value
       end
 
-      def all
-        @value = { :match_all => {} }
+      def all(match_all_args = {})
+        @value = { :match_all => match_all_args }
         @value
       end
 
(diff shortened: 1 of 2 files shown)


woahdae:
Note, this turned out to not be as useful as I had originally thought. The confusion over what exactly norms_field does can be seen [here](https://github.com/elasticsearch/elasticsearch/issues/1019), and for my purposes these arguments didn't help much.

Still a valid patch, but FYI I'm not actually depending on it anymore.

karmi:
@woahdae Errr, sorry for the delay -- it absolutely makes sense to allow passing arguments, thanks for that.

Changed the argument name to `options` to be more consistent, and also used the _boost_ example from elasticsearch's docs, pushed.

Issue #341, opened 2012-05-10
Rails 2.3


loe:
I completely understand how tire is targeted at Rails 3 and beyond, thats is the future.

For those of us with some older apps on 2.3 and new things on Rails 3 running shared infrastructure, it would be handy to have Tire function with 2.3.

I have a fork (https://github.com/loe/tire) that works with 99% of the features. I made two important changes, I commented out the need for active_model, and I provided the rack http patch as seen here: https://github.com/loe/tire/commit/53d851098c86ba210a6b9d4168e65693941ce51b

I'd like to contribute this back to the main project so others can benefit, but I don't like messing with other people's gemspecs. I think ActiveModel should not be required (say like in a Sinatra App) but rather detected and the "additional" model code can be turned on with it.

Thoughts?

karmi:
Hmm, yeah, I think _ActiveModel_ could be ripped out since that's mostly for `Tire::Persistence`, I never focused on making _Tire_ _Rails 2.3_ compatible, but maybe it could be done.

chendo:
+1 on 2.3 support. I was going to use Tire without using the ActiveModel stuff, but the Gemfile requiring activesupport means I have to use @loe's fork. It should be in another gem or something.

TwP:
+1 on this. 

Two gems would be highly recommended. The core DSL for interacting with ElasticSearch should be in the `tire` gem. A second gem needs to be created - `tire-model` perhaps - that contains all the ActiveModel functionality.

@loe thanks for doing all the hard work on this!

TwP:
@loe you should turn this issue into a pull request.

karmi:
Yes, splitting the gems in `core`, `model`, etc does make sense -- provided we keep the easy setup as is the case now. It should be doable over the coming months.

kidpollo:
@loe @karmi has there been any updates in this issue? I would really like to help out. I am currently testing out @loe changes. I have not gotten to deep into the code so far bui it does seem dunctionality can be broken down. When I built https://github.com/kidpollo/tanker I really wanted to make it orm agnostic. Maybe you want to check out some approaches there.

karmi:
@kidpollo No updates yet -- the real solution is to extract the `Tire::Persistence` into separate library, so we don't have to pull ActiveModel directly. I'm not sure how we'll solve the `Tire::Results::Item` compatibility with Rails without ActiveModel though?

loe:
[...]

PR #426, opened 2012-08-07
Remove ActiveModel Dependencies

TwP:
Karel,

Thank you for writing the Tire gem. We are currently using it here at GitHub to interface with several ElasticSearch indexes, and it has made ES much easier to use for us.

Our main Rails application is still running Rails 2.3, though. In order to use Tire in this context we have created a [fork of the project](https://github.com/github/tire) where all the ActiveModel code has been removed. We are contenting to make changes to suit our needs in this fork; however, I do not want this fork to diverge so far from Tire that it essentially becomes a separate project.

Issue #341 describes the very same problem that we encountered. Please let me know how I can help out in the development and improvement of the Tire gem.

I understand that this PR will dramatically change how Tire is used. It is also incomplete in that a second gem needs to be created that encapsulates the ActiveModel code and makes it available to those developers who need it.


karmi:
Hi, thanks for the kind words, Tim -- I'm happy the library is useful at GitHub!

I believe we already touched this issue previously. I agree that Rails 2.x support is highly desirable.

I'd like to tackle this issue by doing a more invasive decomposition/extraction in the Tire codebase, though. Basically, I'd like to split the gem into several inter-related pieces: HTTP client/adapter, raw API matching elasticsearch's API, Ruby DSL on top of that, ActiveModel model integration on top of that, persistence layer on top of that, etc etc.

This should allow to use the API+DSL layer in any Ruby/Rails codebase, and the ActiveModel integration where possible.

Regarding timeline for this refactoring, I think it should be doable within two or three months. Is that feasible for your use case?

karmi:
To all interested: I would very much like the feedback on following questions:
- How to make `Item` usable within Rails without ActiveModel? As soon as you do some `require 'tire/model'`, you need it.
- How to get rid of dependency on ActiveSupport? I hate this dependency as much as any other guy, but the problem is that eg. `Hash#to_param` is very convenient.
- Why support integration tests on Ruby 1.8? Need a really convincing argument, there's a slim chance I'd vote for that.

TwP:
_How to make `Item` usable within Rails without ActiveModel? As soon as you do some `require 'tire/model'`, you need it._

We are not using `Item` at all with Tire at the moment; so this question is a bit of a moot point. There are actually better solutions to this problem in the form of [intridea/hashie](https://github.com/intridea/hashie) and [ahoward/map](https://github.com/ahoward/map).

All of the `tire/model` functionality should be in it's own separate gem. This involves splitting out the basic connectivity and request/response framework into a gem, and then build upon that basic framework the more advanced features such as the query DSL and ActiveModel support.

---

_How to get rid of dependency on ActiveSupport? I hate this dependency as much as any other guy, but the problem is that eg. `Hash#to_param` is very convenient._

The `to_param` method can be implemented in roughly 20 to 30 lines of code. The same can be said of all the ActiveSupport niceties that Tire might need. So steal what is needed, and leave all the other unneeded things behind.

---

_Why support integration tests on Ruby 1.8? Need a really convincing argument, there's a slim chance I'd vote for that._

Tire is definitely your project, Karel. The production environment we are working with is Ruby 1.8.7 at the moment. The solution we need today has to support 1.8.7. That is the _convincing argument_ for me.

Let's flip this argument on its head. Is there any compelling reason to abandon support for 1.8.7? Is there critical functionality that can only be implemented in Ruby 1.9?

If Tire is going to split into two gems (a base functionality layer and then advanced DSL and ActiveModel features as a separate gem), then keeping the base functionality compatible with 1.8.7 is not a hardship. The advanced features can be 1.9 only.

karmi:
@TwP Thanks for your input -- though I don't think you understood well what I have been in fact asking.

PR #430, opened 2012-08-08
Allow classes to be passed instead of index name strings

davekrupinski:
Added functionality to allow classes responding to an :index_name method to be passed into search.  Very useful when defining global or individual index names or prefixes.

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -13,6 +13,7 @@ module Tire
         else
           @indices = Array(indices)
         end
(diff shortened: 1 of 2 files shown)


karmi:
Hey, what's the motivation here? I mean, using `Search::Search.new(ActiveModelArticle.index_name)` is equally easy?

davekrupinski:
It is definitely just an added convenience to the library, but does bring the interface more in line with other common search gems (sunspot, thinking_sphinx, etc).

karmi:
Maybe, but sorry, thins feels too heavy to me...

PR #431, opened 2012-08-13
Adds has_child support for queries

kofno:


karmi:
Hey, good idea, but I think we need a very good integration test as a demonstration of the feature (same as with `nested` queries...)

PR #443, opened 2012-08-29
Add more like this query

AlphaB:


ches:
This exists in https://github.com/karmi/tire-contrib

karmi:
Yes, implemented in https://github.com/karmi/tire-contrib/blob/master/lib/tire/queries/more_like_this/more_like_this.rb. Thanks!!

PR #445, opened 2012-08-31
Boosting Query support added

martinciu:
Boosting Query support: http://www.elasticsearch.org/guide/reference/query-dsl/boosting-query.html


karmi:
Marcin, thanks!, very nice patch, cleaned up the integration test a bit and merged.

PR #446, opened 2012-09-02
Acceptig symbol as :as parameters when indexing and documents and adding documentation on how to use lambda's on it

mauricio:
Hello guys,

This PR contains code to accept `Symbol`s and `lambda`s as as values for `:as` options when indexing. It contains tests covering this new code and also tests for the current implementation.

I have added validation for when `:as` fields are set to make sure that if the user gives it an unknown or invalid value an exception will be raised instead of failing silently (or breaking as in the case of `lambda`s).

I have also added information to the README file pertaining these changes when indexing documents.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -440,7 +440,37 @@ You can define different [_analyzers_](http://www.elasticsearch.org/guide/refere
 or any other configuration for _elasticsearch_.
 
 You're not limited to 1:1 mapping between your model properties and the serialized document. With the `:as` option,
-you can pass a string or a _Proc_ object which is evaluated in the instance context (see the `content_size` property).
+you can pass a symbol, string, _lambda_ or a _Proc_ object to generate the value. The evaluation is done as follows:
+
+If `String` the text given is evaluated at the current instance (using `instance_eval`), look at the `content.size` property above.
+
+If `Symbol` a method with the same name as the symbol is called at the current instance, as in:
+
+```ruby
+    mapping do 
+     indexes :id, :as => :some_other_id
+    end
+    
+    def some_other_id
+      self.component.id
+    end    
+```
+
+If `Proc`, the proc is evaluated against the current instance (using `instance_eval`), as in:
+
+```ruby
+    mapping do 
+     indexes :author_name, :as => proc { author.name }
+    end
+```
+
+If `lambda`, the lambda is evaluated passing the current object as a parameter:
+
+```ruby
+    mapping do 
+     indexes :author_name, :as => lambda { |article| article.author.name }
+    end
+```
 
 Chances are, you want to declare also a custom _settings_ for the index, such as set the number of shards,
 replicas, or create elaborate analyzer chains, such as the hipster's choice: [_ngrams_](https://gist.github.com/1160430).
--- a/README.markdown
+++ b/README.markdown
@@ -447,13 +447,13 @@ If `String` the text given is evaluated at the current instance (using `instance
 If `Symbol` a method with the same name as the symbol is called at the current instance, as in:
 
 ```ruby
-    mapping do 
-     indexes :id, :as => :some_other_id
+    mapping do
+      indexes :category, :as => :category_name
     end
-    
-    def some_other_id
-      self.component.id
-    end    
+
+    def category_name
+      self.category.name
+    end
 ```
 
 If `Proc`, the proc is evaluated against the current instance (using `instance_eval`), as in:
--- a/lib/tire/invalid_as_option_exception.rb
+++ b/lib/tire/invalid_as_option_exception.rb
@@ -1,0 +1,9 @@
+module Tire
+  class InvalidAsOptionException < Exception
+
+    def initialize( value )
+      super("Unknown type for :as field, accepted types are `Symbol`, `String` and `Proc`s - #{value.class.name} - #{value.inspect}")
+    end
+
+  end
+end
--- a/lib/tire/invalid_as_option_exception.rb
+++ b/lib/tire/invalid_as_option_exception.rb
@@ -1,9 +1,0 @@
-module Tire
-  class InvalidAsOptionException < Exception
-
-    def initialize( value )
-      super("Unknown type for :as field, accepted types are `Symbol`, `String` and `Proc`s - #{value.class.name} - #{value.inspect}")
-    end
-
-  end
-end
--- a/lib/tire/model/indexing.rb
+++ b/lib/tire/model/indexing.rb
@@ -78,6 +78,9 @@ module Tire
         # * Use the `:as` option to dynamically define the serialized property value, eg:
         #
         #       :as => 'content.split(/\W/).length'
+        #       :as => :author_name                        # calls the author_name method
+        #       :as => proc { category.name }              # evaluates the proc at the current instance
+        #       :as => lambda { |item| item.product.name } # evaluates the lambda passing the current instance as a parameter
         #
         # Please refer to the
         # [_mapping_ documentation](http://www.elasticsearch.org/guide/reference/mapping/index.html)
@@ -85,6 +88,15 @@ module Tire
         #
         def indexes(name, options = {}, &block)
           mapping[name] = options
+
+          unless options[:as].nil?
+            case options[:as]
+              when String, Symbol, Proc
+                true # ok, don't do anything
+              else
+                raise Tire::InvalidAsOptionException.new( options[:as] )
+            end
+          end
 
           if block_given?
             mapping[name][:type]       ||= 'object'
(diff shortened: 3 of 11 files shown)


mauricio:
Anyone? @karmi

karmi:
Hi, sorry for the delay, busy lately. I definitely like the direction and the behaviour, but would change something in the implementation a bit... Allow me couple of days to look at it...

lib/tire/model/indexing.rb
```diff
@@ -86,6 +89,15 @@ def mapping(*args)
         def indexes(name, options = {}, &block)
           mapping[name] = options
 
+          unless options[:as].nil?
+            case options[:as]
+              when String, Symbol, Proc
+                true # ok, don't do anything
+              else
+                raise Tire::InvalidAsOptionException.new( options[:as] )
```
karmi, line 97: Why raise here -- why not in the `to_indexed_hash` where we are doing the `case` anyway?

lib/tire/invalid_as_option_exception.rb
```diff
@@ -0,0 +1,9 @@
+module Tire
+  class InvalidAsOptionException < Exception
+
+    def initialize( value )
+      super("Unknown type for :as field, accepted types are `Symbol`, `String` and `Proc`s - #{value.class.name} - #{value.inspect}")
+    end
+
+  end
+end
```
karmi: 1/ Beware of files without trailing newlines in git.
2/ I think we should have a file like `errors` or something, rather then create a new file for every exception there is.

README.markdown
```diff
@@ -440,7 +440,37 @@ You can define different [_analyzers_](http://www.elasticsearch.org/guide/refere
...
-you can pass a string or a _Proc_ object which is evaluated in the instance context (see the `content_size` property).
+you can pass a symbol, string, _lambda_ or a _Proc_ object to generate the value. The evaluation is done as follows:
+
+If `String` the text given is evaluated at the current instance (using `instance_eval`), look at the `content.size` property above.
+
+If `Symbol` a method with the same name as the symbol is called at the current instance, as in:
+
+```ruby
+    mapping do 
+     indexes :id, :as => :some_other_id
```
karmi: Thanks for having added also docs for it, that's rare :) But I think we should try for a better example? I usually hate "foobar", "someotherid" in documentation...

test/unit/model_search_test.rb
```diff
@@ -907,6 +907,28 @@ def self.find(*args); new; end
...
+        should 'raise a Tire::InvalidAsOptionException if :as is of an invalid type' do
+
+          assert_raise ::Tire::InvalidAsOptionException do
+
+            class SomeFakeModel
+              include Tire::Model::Search
+              include Tire::Model::Callbacks
+
+              mapping do
+                indexes :name, :as => Time.now
```
karmi, line 921: I'm trying to figure out if there isn't a valid use case for passing an object like this... Some application-level configuration value, which is determined at application startup?

lib/tire/model/indexing.rb
mauricio, line 97: To avoid allowing you to define a config that's not supported, just failing early. But I can move this to the to_indexed_hash method too.

test/unit/model_search_test.rb
mauricio, line 921: You could possibly use it for something like that, but you could also use a `proc/lambda` for it, my main concert here is to only accept something that we know how to handle. 

When I first used `:as` I was using it with a `lambda` and it was working, somewhere along the path it stopped working and I was clueless as to what I was missing (which was my own mistake, as I didn't notice the docs did not talk about lambdas, only procs), that's when I decided to check the code :)

So, this is mostly me scratching my own itch and making sure it is not possible to send in a value that's not "known".

Now that's for you to decide if we should validate or just keep it as is ;D (i'm personally much more inclined to validate)

mauricio:
I have updated the README and also added an errors.rb file.

test/unit/model_search_test.rb
karmi, line 921: I understand your reasoning. Anybody else has any opinion on this?

mauricio:
No other comments on this @karmi ?

kenshin54:
+1

test/unit/model_search_test.rb
ches, line 921: It seems sensible to me. If someone came along with a good use case, it'd probably have to be something where a given object would need to `respond_to?` some common convention, right? I don't know how else you could sensibly handle it, really. In that case it'd be a simple matter to add a `when` clause to the case statement @mauricio added to support that.

I like the clarification/definition of the semantics here otherwise.

karmi, line 921: @ches Don't understand your reasoning 100%, I'm afraid, but still -- it's really the same thing as defining _scopes_ in ActiveRecord, see eg. http://paulbarry.com/articles/2008/12/18/named-scope-to-lambda-or-not-to-lambda-that-is-the-question

ches, line 921: Can you give an example of the scope-like usage you're referring to?

karmi, line 921: @ches, in the article above: `named_scope :published_before, lambda{ |date|  {:conditions => ["published_at < ?", date] }  }`

ches, line 921: @karmi I understand ActiveRecord scopes, I was asking for an example of how you imagine that being applied in the case of `indexes :as` :smile:

karmi, line 921: Like this:

``` ruby
require 'tire'

Tire.index('articles').delete

class Article

  property :title
  property :size,       :as => proc { title.size }
  property :created_at, :default => lambda { Time.now }

end


p Article.create(title: 'Hello').attributes,

  Article.index.refresh,

  "Size: %s" % [ Article.first.size ]
```

karmi, line 921: Or, more to the point:

``` ruby
require 'tire'

Tire.index('articles').delete

class Article

  property :title
  property :mydate,     :as => proc { Time.now }
  property :created_at, :default => lambda { Time.now }
end

p "Current date: %s" % [ Time.now ]
sleep 5

Article.create
Article.index.refresh

p "My date:      %s" % [ Article.first.mydate ]
```

ches, line 921: But shouldn't that work as-is? You're using `proc`/`lambda`, the same way we naturally would when defining AR scopes like this, and I don't think that is at question here. Your initial question was if there might be a use case for passing some arbitrary object for `:as` -- that's what I'm asking, for a (hypothetical) example of how you could imagine that being used, sorry if that was unclear. I'm not aware of any analog to that in AR scope definitions.

@mauricio hasn't changed anything about Tire's existing ability to handle procs/lambdas here, he just wants to bail if given something else that we don't know what to do with. That seems sane and doesn't lock you out of later accepting something else if you came up with something that would be meaningful.

karmi:
Guys, see both commits added to this issue:
- https://github.com/karmi/tire/commit/0ccee42ccfc1bba9c4264a8e9c56c3b4965cfb49
- https://github.com/karmi/tire/commit/bbbf372a30e484856435ff50e73aaffeb6669a1d

I'm leaning towards the latter. It is, though, based only on a gut feeling, not some hard-held opinion; I'd rather keep the behaviour open ended then be too strict about it.

/cc @vhyza

karmi:
Closing, going with karmi/tire@bbbf372 for the time being.

ches:
Regardless of the decision on the handling of arbitrary objects that was discussed here, I still think the documentation @mauricio contributed was helpful -- I was just looking for similar myself (not the first time I've needed a reference!), and after going to code where it isn't obvious, I remembered this pull request.

karmi:
@ches I'm afraid I don't understand what you want Tire to do here. The behaviour is pretty expectable -- when you put `as: Time.now`, you're "freezing" `Time.now` and that's it. Of course, it should be properly documented in the code -- but definitely not in the README.

ches:
Oh, actually I didn't realize until a second look just now that the possibility of giving a Symbol as an instance method reference was not actually accepted -- otherwise I was going to say there's nothing additional I'm looking for from Tire, just the docs. The Symbol support is a small nicety though IMO, it's a common pattern and a little cleaner than thinking about `instance_eval` any time it isn't absolutely necessary.

Where would you want such documentation to be? I assume expanding the explanation of `:as` in the RDoc for `indexes` on `Tire::Model::Indexing`?

ches:
(If you look at that method, and you're not familiar with Tire's codebase more fully, it's not at all apparent where `:as` actually gets handled -- that's why I'm being vocal about a wish for more docs).

PR #449, opened 2012-09-06
abstracts bulk api handling to a new method 

rhec:
adds new methods to access the delete and create features of the bulk api
I've been using the bulk delete in production for a while with no problems.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -84,8 +84,8 @@ module Tire
       curl = %Q|curl -X POST "#{url}" -d '#{document}'|
       logged([type, id].join('/'), curl)
     end
-
-    def bulk_store(documents, options={})
+    
+    def bulk_api_action(action, documents, options={})
       payload = documents.map do |document|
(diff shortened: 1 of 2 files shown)


karmi:
Full changelog: https://github.com/karmi/tire/compare/8e81c3a...1c0d043

karmi:
Rob — thanks, finally merged and cleaned up & amended up a bit, released as 0.5.1.

woahdae:
Hey, thanks for adding support for parent relationships, but am I crazy, or is this weird: To get parent/child behavior I need to put the parent option in every document in ruby, so it can be subsequently deleted from the documents and inserted into the bulk meta, which just seems odd. It's not like you can have multiple parents per bulk action, so why not just pull it from the mapping? Plus #store accepts parent, but #update_index doesn't support it?

Maybe it's particular to my use case, but I would have expected to put :_parent in the mapping, since ES needs that already, and have #bulk and #store pull parent/routing info from the mapping.

I have a monkeypatch that does this already, prior to these changes, and am willing to make a pull request if you think it's a sane strategy.

PR #451, opened 2012-09-09
add test helpers for test/unit and rspec, update readme

jalkoby:
Add test helpers for Test::Unit and Rspec, update Readme to new changes


karmi:
I'd prefer to add a support like that into the https://github.com/karmi/tire-contrib gem, not into the gem directly. What do you think?

While some out of the box support is definitely nice, people's use cases wary a lot and it's trivial to do setup/teardown manually..

mereghost:
+1 (on tire-contrib)

Seems like a prime candidate for tire-contrib (and I'd welcome it) =)

jalkoby:
Maybe you are right. So I close this pull request. See you in tire-contrib)

PR #452, opened 2012-09-12
Add --data-binary as curl switch for logs when bulk loading (as per docs)

canadaduane:
The curl output should use --data-binary instead of -d when describing the action in the Tire log output. See http://www.elasticsearch.org/guide/reference/api/bulk.html.

Also adds count of omitted documents in log output.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -117,7 +117,7 @@ module Tire
         end
 
       ensure
-        curl = %Q|curl -X POST "#{url}/_bulk" -d '{... data omitted ...}'|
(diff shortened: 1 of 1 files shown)


karmi:
Well, I think the docs say that you should use the option when you're providing the payload as `@file`. Nevertheless, the curl log is just informational as it is -- otherwise it would dump _loads_ of payload data into your log... Leaning towards closing this?

canadaduane:
Well, it just tripped me up when trying to debug. It would be a time saver
in rare cases, if it didn't slightly mislead.

karmi:
OK, thanks, merged.

PR #468, opened 2012-10-04
Added additional sort example

mdpatrick:
Very prominent feature of elasticsearch is that it allows multi-field sorting. Added it to readme for tire.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -409,7 +409,7 @@ advanced facet aggregation, highlighting, etc:
     Article.search do
       query             { string 'love' }
       facet('timeline') { date   :published_on, :interval => 'month' }
-      sort              { by     :published_on, 'desc' }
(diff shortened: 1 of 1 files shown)


karmi:
Yes, _maybe_, but definitely without the comment... Still on the fence with this. We don't want to explain everything about elasticsearch in the README, do we?

karmi:
Should be really properly documented in https://github.com/karmi/tire/blob/master/lib/tire/search/sort.rb and https://github.com/karmi/tire/blob/master/lib/tire/search.rb#L60-63 ...

mdpatrick:
Would you be OK with me re-submitting the PR with the change I made less the comment?

karmi:
@mdpatrick Absolutely, if you can add the documentation into those places (preferably in YARDoc format, see eg. https://github.com/karmi/tire/blob/master/lib/tire/alias.rb)!

I'm sorry but I don't wanna stretch the README too far -- some people have trouble reading it as it is :)

PR #470, opened 2012-10-07
Specify a version of minitest that is compatible with mocha

carols10cents:
This should fix the current Travis failures.

### DIFF
--- a/tire.gemspec
+++ b/tire.gemspec
@@ -45,7 +45,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "bson_ext"
   s.add_development_dependency "redis-persistence"
   s.add_development_dependency "curb"
-  s.add_development_dependency "minitest"
+  s.add_development_dependency "minitest", "~> 2.12"
   s.add_development_dependency "turn", "~> 0.9" if defined?(RUBY_VERSION) && RUBY_VERSION > '1.9'
 
   # These gems are not needed for CI at <http://travis-ci.org/#!/karmi/tire>


karmi:
Hi, thanks -- any idea why it fails on TravisCI without the version? Can you ping them? I'd like to stay as far from gem versions as much as possible (even though this is only for development).

carols10cents:
Does it not fail for you locally? What if you delete your Gemfile.lock, bundle, then run the tests? Since the Gemfile.lock isn't checked in and since a version isn't specified, Travis is getting the latest version of all the gems-- but you might have a local Gemfile.lock that has earlier versions.

The specific error I see [on travis build 311.1](https://travis-ci.org/#!/karmi/tire/jobs/2704520) is:

```
/home/travis/.rvm/gems/ruby-1.9.3-p194/gems/mocha-0.12.6/lib/mocha/integration/mini_test.rb:56:in 
`<class:TestCase>': No Mocha monkey-patch for MiniTest version (RuntimeError)
```

There's been some work/discussion around this:

https://github.com/freerange/mocha/issues/87
https://gist.github.com/3137094

and the latest code in mocha's github repo works fine if I specify that in the Gemfile.

But it still looks like mocha 0.12.6 and minitest 4.1.0 are incompatible, so I don't think there's a way around this other than waiting for a new version of mocha to be released, using mocha's latest from github, or locking down the minitest version at something earlier than 4.

karmi:
> Does it not fail for you locally? What if you delete your Gemfile.lock, bundle, then run the tests? 

No. When I delete `Gemfile.lock`, delete all `minitest` and `mocha` gems, run `bundle install` and then run `bundle exec rake test`, everything runs smoothly. Could you run it locally on a fresh checkout, with gems deleted, etc?

carols10cents:
[Looks like mocha released 0.12.7](https://travis-ci.org/#!/karmi/tire/jobs/2779445/L45), so that's why it's working for you now (and passing on travis). Hopefully that'll be the end of this issue....

karmi:
@carols10cents Cool. Prefer to sort it that way then to pin gem versions -- been burned with gems doing that too often in the past...

karmi:
Hi, got bit by the error just today. Pulled in your fix, to be on the safe side with Mocha and Minitest.

PR #472, opened 2012-10-09
Implement partial fields

uriagassi:


karmi:
Thanks!, cleaned up a bit and merged.

PR #473, opened 2012-10-09
implement respond_to? for Item

slawosz:


karmi:
Thanks, scandalous oversight on my part .)

PR #478, opened 2012-10-11
Fix lambda error in README.markdown

bgadoury:
The two lambda examples raise this: ArgumentError: wrong number of arguments (1 for 0)


karmi:
Argh, of course, you're absolutely right, merged+pushed, thanks!

PR #480, opened 2012-10-13
Fixing elapsed time presentation on import task

wandenberg:
The time was being shown in a wrong way when the elapsed time is a float number.

### DIFF
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -17,23 +17,6 @@ namespace :tire do
   DESC
   desc full_comment_import
   task :import do |t|
-
-    def elapsed_to_human(elapsed)
-      hour = 60*60
-      day  = hour*24
-
-      case elapsed
-      when 0..59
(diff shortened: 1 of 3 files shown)


karmi:
Hi, thanks for looking into that so thoroughly! I consider the Rake task just a utility method, though, and don't think we should add some `Utils` namespace just for that...

wandenberg:
Hi,
I moved the code from tasks.rb to utils.rb to be easier test it.
If you want I can apply the fix on tasks.rb only, just will stay without tests.
Can I do that? The fix is simple but a good presentation be part of a good software.

Regards,
Wandenberg

karmi:
Hi, you're absolutely right it should have been fixed. I just fixed the method in `tasks.rb`, see karmi/tire@6fef487.

While the import Rake task is very useful, it's not a end-all solution and people should be more then ready to write their own feeder scripts; it's very, very convenient and easy to do so in Ruby. The Rake task is a mere convenience for simple cases and a help for beginners.

PR #453, opened 2012-10-18
finders are not thread safe

grantr:
The methods defined in Tire::Model::Persistence::Finders are not thread-safe. They set the wrapper globally and expect it to be the same later. Concurrent finds attempt to wrap documents with the wrong wrapper.

This could be fixed by passing the :wrapper argument to search and retrieve from the finders instead of setting it globally. Index#retrieve would need to take a :wrapper option.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -184,7 +184,7 @@ module Tire
       logged(id, curl)
     end
 
-    def retrieve(type, id)
+    def retrieve(type, id, options={})
       raise ArgumentError, "Please pass a document ID" unless id
 
       type      = Utils.escape(type)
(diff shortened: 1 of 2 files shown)


karmi:
I don't think it makes sense to fiddle with injecting wrapper to `search` at the moment... Is your concern academic or are you seeing practical problems with wrapper being not thread safe? Could [`Thread.current`](http://ruby-doc.org/core-1.9.3/Thread.html#method-i-5B-5D) hash be used to mitigate the issue?

grantr:
I have actually seen this in very light testing of multithreaded apps. 

The change I'm suggesting is a single line in Index#retrieve and a couple lines in finders. Using Thread.current would be a much bigger and far-reaching change. Tire::Search already takes a :wrapper option (via Results::Collection), so it makes sense for Index#retrieve to do the same and for the finders to use that option.

karmi:
@grantr Apologies -- didn't realized there was a patch attached while solving many issues at the same time.

This implementation is much cleaner then the original mess. Thanks!, merged.

grantr:
There was no patch originally but I added one. I already had the code in my own project so I figured I might as well submit it.

Thanks for merging!

karmi:
Cool, thanks!

PR #492, opened 2012-10-24
Tire.indices and Tire.total_storage methods

fabn:
I've used those methods in a project, you can merge them if you think they could be useful to others


karmi:
Hi, thanks, but I think something like this needs a much more thorough approach... You should put it out as a gist and blog about it :)

PR #496, opened 2012-10-29
Using Tire::search without payload produces unexpected results

timon:
Running this piece of code 

``` ruby
  s = Tire.search 'articles-test'
  s.query { string 'ruby' }
  s.facet 'tags' do
      filter :term, :tags => 'ruby'
  end
```

produces full index contents in result.

As I take it from documentation, building query step by step should be possible.
This pull request allows building query without specifying payload on Tire::search call

### DIFF
--- a/lib/tire/dsl.rb
+++ b/lib/tire/dsl.rb
@@ -10,6 +10,8 @@ module Tire
         Search::Search.new(indices, options, &block)
       else
         payload = case options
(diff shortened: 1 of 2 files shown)


karmi:
Thanks for the catch!, did it differently, but agreed this makes sense to allow...

timon:
Thanks!

PR #497, opened 2012-10-29
:recreate_elasticsearch_index method for model indexing

rsmdt:
I find this defining in my projects, which I think is handy for testing.

Best,
Rudi


karmi:
Sorry, but I think everybody can just put those two lines in their setup/teardown...

PR #517, opened 2012-11-15
Use an instance method as after_save/destroy parameter instead of a lambda.

pi3r:
In some cases (unit testing), i don't want my models to perform any Tire callbacks. Now that the callback method is named, i can do the following:

``` ruby
before(:all) do
  Model.skip_callback :save, :after, :_tire_update_index_callback
end
```

I didn't write any tests because i don't change the behavior,  `update_index` is still called.

### DIFF
--- a/lib/tire/model/callbacks.rb
+++ b/lib/tire/model/callbacks.rb
@@ -18,8 +18,8 @@ module Tire
         # Update index on model instance change or destroy.
         #
         if base.respond_to?(:after_save) && base.respond_to?(:after_destroy)
-          base.send :after_save,    lambda { tire.update_index }
-          base.send :after_destroy, lambda { tire.update_index }
+          base.send :after_save,    :_tire_update_index_callback
+          base.send :after_destroy, :_tire_update_index_callback
         end
 
         # Add neccessary infrastructure for the model, when missing in
@@ -34,6 +34,9 @@ module Tire
 
       end
 
+      def _tire_update_index_callback
+        tire.update_index
+      end
     end
 
   end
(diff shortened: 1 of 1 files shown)


karmi:
Hi, I understand the intention, but isn't it better to [stub](http://rubydoc.info/gems/rspec-mocks/RSpec/Mocks/Methods:stub) the callback here, based on the fact that you _are_ actually executing it in the code?

Also, I'm not a fan of polluting the namespace any more then is neccessary...

pi3r:
Hello, i understand that `stub` is better but, for example, when i have the following factory:

``` ruby
# 
factory :product do
  ...
  after(:create) do |product|
    product.class.tire.index.refresh
  end
end
```

How can i know if i need to refresh the index or not ? With a basic ugly piece of code i could do:

``` ruby
if Product._save_callbacks.select{ |k| k.raw_filter == :_tire_update_index_callback }.present?
  product.class.tire.index.refresh
end
```

The main point for me is that using a method instead of a lambda gives visiblity.
Example: 
When i call Model._save_callbacks (Callbacks as lambda), i get something like :

```
 @raw_filter=
   #<Proc:0x00000005b67718@/home/pierre/.rbenv/versions/1.9.3-p125/lib/ruby/gems/1.9.1/gems/tire-0.5.1/lib/tire/model/callbacks.rb:21 (lambda)>>
```

But with the callbacks as method, i get something like:

```
 @raw_filter=:_tire_update_index_callback>
```

We can see instantly the purpose of the callback. 

Cheers

karmi:
> How can i know if i need to refresh the index or not?

I'm sorry, but I don't understand 100% what are you trying to do here: if the model has Tire integration, and you have an integration test, you just need to refresh the index so data inserted in test setup is visible. For unit tests, just use the `build`, not the `create` strategy?

PR #518, opened 2012-11-15
Add Travis png to the readme

pi3r:
Now we can see the build status in the readme :)

Preview here -> https://github.com/pi3r/tire/blob/readme/README.markdown


karmi:
I'm all for it once the Travis status images are not as ugly as they currently are...

PR #520, opened 2012-11-15
Put/Delete mapping support

cjbottaro:
My put_mapping branch was branched off of the v0.5.1 tag, but it merges cleanly into master (as of 11:45 am today).  I added a couple integration tests and one unit test.  All tests pass (both on the branch and on master after merging it).


psykidellic:
I just started using ES/Tire and during development/staging to have this feature as we go on changing the mappings, as we go.

reiz:
An example in the documentation would be useful. I need to know how I can send a mapping via PUT to an existing empty index. 

This is an issue for me right now. There is one ES SaaS provider I found, which is creating the index for me. And I am not allowed to delete or overwrite it. I am forced to send my mappings via a PUT to an existing index. Right now Tire is not usable for this task. 

I appreciate any kind of feedback.

karmi:
Finally found time to evaluate and refactor the patch, sorry for the delay! Merged & pushed.

PR #521, opened 2012-11-15
Add count to index

brandon-beacher:
Hello - I've added support for _count on an index to retrieve the count of documents in the index.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -52,6 +52,11 @@ module Tire
       alias_name ? Alias.all(@name).select { |a| a.name == alias_name }.first : Alias.all(@name)
     end
 
+    def count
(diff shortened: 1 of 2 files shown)


karmi:
Hi, thanks! Two comments:

1/ We should support all the options for the [_Count API_](http://www.elasticsearch.org/guide/reference/api/count.html), notably the ability to pass a query. 

2/ The count is already supported, via the `search_type` parameter. This is generally preferable in elasticsearch, so maybe `Index#count` should be just a nice wrapper for that?

karmi:
Decided it actually makes more sense to define a separate API/DSL for the [_Count API_](http://www.elasticsearch.org/guide/reference/api/count.html), so complicated it a bit :)

Thanks!, the feature should be compatible with your patch.

PR #532, opened 2012-11-29
Make Results::Item#respond_to? conform to interface

Odaeus:
Ruby's native Object#respond_to? accepts a second boolean parameter to determine if private methods should be included.


karmi:
Thanks, merged & added a test case.

PR #534, opened 2012-12-05
Allow arbitrary ordering of facet methods.

msonnabaum:
Ran into an issue where calling facet_filter before other facet methods would fail:

``` ruby
facet "test" do
  facet_filter :range, { :published_on => { :from => '2011-01-01', :to => '2011-01-01' } }
  histogram :wt, :interval => 70
end
```

Reordering prevents the issue, but that seems a bit brittle to me.

Perhaps this PR is a bit safer?

### DIFF
--- a/lib/tire/search/facet.rb
+++ b/lib/tire/search/facet.rb
@@ -11,52 +11,53 @@ module Tire
       def initialize(name, options={}, &block)
         @name    = name
         @options = options
+        @value = {}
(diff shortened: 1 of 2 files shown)


karmi:
Hello, that seems a valid concern! Could you add a test demonstrating the issue, the way you describe it in the issue?

msonnabaum:
Added test. Passes with the PR, throws an error without.

karmi:
Cool!, added an integration test as well, amended the commit message etc and pushed, thanks!

Issue #529, opened 2012-11-24
Bulk import of Mongoid documents incorrect on Tire 0.5.1


digitalplaywright:
Bulk import of Mongoid documents in Tire 0.5.1 creates documents in the index with an erroneous id. The document id in the ES index is assigned a random string id instead of the bson object id. The same error may also be present on new document creation, but it seems to happen quite seldomly and  we do not have a testcase that demonstrates this. 

This example demonstrates the problem:

```
class TestCase
  include Mongoid::Document
  include Tire::Model::Search
  include Tire::Model::Callbacks

  cattr_reader :per_page
  @@per_page = 10

  def self.paginate(options = {})
    page(options[:page]).per(options[:per_page])
  end

  def self.build_new
    TestCase.destroy_all
    TestCase.tire.index.delete

    for i in 1..100
      tk = TestCase.new(:name => "foobar")
      tk.save!
    end

    TestCase.tire.index.delete
    TestCase.create_elasticsearch_index
    TestCase.tire.import
  end

  field :name, :type => String

  mapping do
    indexes :name,  :type => 'string'
  end

end
```

We use Mongoid 3.0.1, Moped 1.3.0 and Origin 1.0.11. We use Elastic Seach 0.19.11. Please let us know if you are able to reproduce the problem.

konsti:
I have the same problem on **ElasticSearch 0.19.11** with:
- Ruby 1.9.3p327
- Rails 3.2.9
- Tire 0.5.1
- MongoId 3.0.13
- Moped 1.2.9
- Origin 1.0.10

digitalplaywright:
@konsti I have not had time to look debug Tire to figure out what is going wrong here. Have you made any progress on this issue?

@karmi Please let us know if you have any advice on what may be the culprit code in Tire or if you have time to help debug this issue.

konsti:
@digitalplaywright We use this versions in a small side project, so I just don't use bulk import. Didn't find any time to investigate further. I'll keep you posted.

sideshow:
See this pull request karmi/tire#535, which should fix this issue.

karmi:
It seems that _Mongoid_ returns a BSON object, instead of some plain JSON-friendly object for the IDs. See tire#535.

Either we're incorrectly serializing _Mongoid_ objects in _Tire_, or _Mongoid_ needs some special care. Any _Mongoid_ specialist to advice here? //cc @michaelklishin

digitalplaywright:
[...]

PR #535, opened 2012-12-05
Fixes issue 529 - bulk importing mongoid documents gives an erroneous id

sideshow:
Document#id with Mongoid returns a Moped::BSON::ObjectId instead if a string, which was confusing elasticsearch. Casting the id to a string fixes this issue.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -415,7 +415,7 @@ module Tire
         when document.is_a?(Hash)
           document[:_id] || document['_id'] || document[:id] || document['id']
         when document.respond_to?(:id) && document.id != document.object_id
-          document.id
+          document.id.to_s
       end
       $VERBOSE = old_verbose
       id


digitalplaywright:
This bugfix seems to fix the problem. Thanks!

karmi:
Hi, there're good intentions in this commit, but I don't think it's that easy, and it's impossible to merge as it is.
1. We're converting the `id` attribute of documents people send into _elasticsearch_ into a `String`. That sounds dangerous.
2. That _Mongoid_ returns an BSON object instead of some JSON friendly object is quite unfortunate. I'm all for making Tire work smoothly with Mongoid, but not with inconsistent fixes like this.

sideshow:
Good point. The BSON Object does include a JSON friendly object, but its in a format that is good for mongo and not for elasticsearch.  `{ "$oid": "ABC..."}`. Perhaps it would be better to use `document.id.is_a?(Moped::BSON::ObjectId)` ?

karmi:
Closing in favor of karmi/tire#529.

PR #555, opened 2012-12-20
JRuby compatibility

dwbutler:
I'm thinking of using Tire on a project that will run in JRuby. I noticed the .travis.yml didn't include JRuby, so I  thought I'd try to get the tests running under JRuby. I was successful - the unit and integration tests all run successfully under JRuby.

This patch modifies the gem spec and test framework so it will bundle and run under JRuby. No change to the actual Tire implementation was necessary.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,14 +3,6 @@ source "http://rubygems.org"
 # Specify your gem's dependencies in tire.gemspec
 gemspec
 
-platform :ruby do
-  gem "yajl-ruby",   "~> 1.0"
-  gem "sqlite3"
-  gem "bson_ext"
-  gem "curb"
-  gem "oj"
-end
-
 unless defined?(RUBY_VERSION) && RUBY_VERSION < '1.9'
   gem "turn", "~> 0.9"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,20 @@ source "http://rubygems.org"
 
 # Specify your gem's dependencies in tire.gemspec
 gemspec
+
+platform :ruby do
+  gem "yajl-ruby",   "~> 1.0"
+  gem "sqlite3"
+  gem "bson_ext"
+  gem "curb"
+  gem "oj"
+end
+
+unless defined?(RUBY_VERSION) && RUBY_VERSION < '1.9'
+  gem "turn", "~> 0.9"
+end
+
+platform :jruby do
+  gem "jdbc-sqlite3"
+  gem "activerecord-jdbcsqlite3-adapter"
+end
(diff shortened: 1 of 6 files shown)


karmi:
Hello David, we had support for JRuby at one point of time. In the end, I made the decision all the conditinials sprinkled around the test suite are not worth it -- but your patch looks reasonably clean. Given the current JRuby craze, I think we should support it.

Do you think the supporting the `jruby-18mode` variant is neccessary? I don't like to overload Travis with a big matrix.

dwbutler:
Hi, I figured since you have Travis running the unit tests in 1.8.7, that it would be good to run them in JRuby 1.8 mode as well for completeness sake. Up to you if you want to leave that in or not.

Gemfile
```diff
@@ -2,3 +2,24 @@ source "http://rubygems.org"
 
 # Specify your gem's dependencies in tire.gemspec
 gemspec
+
+platform :ruby do
+  gem "yajl-ruby",   "~> 1.0"
+  gem "sqlite3"
+  gem "bson_ext"
+  gem "curb"
```
karmi: I see you've transferred gem dependencies in the Gemfile -- while it's a nice way to separate the dependencies for platforms, AFAIK it breaks the dependency information at Rubygems itself? Can we support Ruby/JRuby in the .gemspec?

dwbutler:
Unfortunately, I've found it impossible to define platform-dependent dependencies in the .gemspec. There are several reasons for this.
- The most important reason is that the .gemspec format does not have a way to define platform dependencies!
- Currently, an officially released gem on rubygems.org must defined a fixed set of dependencies. If you take a look at https://rubygems.org/gems/tire, you'll see the fixed list of runtime and development dependencies. Running `gem install tire` will install all the listed runtime dependencies. Checking out the repo and running `bundle install` will install everything in the .gemspec, as well as in the Gemfile. This is what currently fails in JRuby.
- You technically can put platform checks in the .gemspec, but this does not have the desired result. What happens is when you `gem build`, the gem will be built in whatever platform you happen to be running on your machine. The published gem spec on rubygems.org will then reflect whatever was built on your local machine.

Doing the platform checks in the Gemfile is the only fool-proof way I know to actually get `bundle install` working, and I've seen this method used in many other gems.

I also think it makes sense to only publish the platform-agnostic gems as development dependencies in the official gem spec on rubygems.org, if JRuby is to be supported.

However, if you wish, you could do the platform checks in the .gemspec, and then make sure to always use MRI to actually release the gem. This would allow all the gems to be listed as development dependencies on rubygems.org.

dwbutler:
@karmi, I made some improvements and implemented some of the suggestions you had (such as putting the development dependencies back in the gemspec). Travis is happy too (https://travis-ci.org/dwbutler/tire/builds/3940134). Let me know what you think!

karmi:
@dwbutler This is good news!, will check it today.

karmi:
@dwbutler Thanks!, merged & pushed with minor cleanups.

PR #558, opened 2012-12-25
Persistence: Fix invalid casting of nil in casted collections

barakcoh:
When using `Tire::Model::Persistence` with a casted collection, casting fails when the value is nil.

Suggested patch will always cast nil value to nil (i.e. do nothing)

**Example**

``` ruby
class Job
  def initialize(params);                      @attributes = HashWithIndifferentAccess.new(params);  end
  def method_missing(method_name, *arguments); @attributes[method_name];                             end
  def as_json(*);                              @attributes;                                          end
end

class Company

  property :name,         :analyzer => 'keyword'  
  property :jobs,         :class    => [Job]
end
```

```
1.9.3p194 :006 > Company.create name: 'foo', jobs: nil
NoMethodError: undefined method `new' for [Job]:Array
    from /Users/barak/.rvm/gems/ruby-1.9.3-p194/gems/tire-0.5.2/lib/tire/model/persistence/attributes.rb:124:in `__cast_value'
    from /Users/barak/.rvm/gems/ruby-1.9.3-p194/gems/tire-0.5.2/lib/tire/model/persistence/attributes.rb:110:in `block in __update_attributes'
    from /Users/barak/.rvm/gems/ruby-1.9.3-p194/gems/tire-0.5.2/lib/tire/model/persistence/attributes.rb:110:in `each'
    from /Users/barak/.rvm/gems/ruby-1.9.3-p194/gems/tire-0.5.2/lib/tire/model/persistence/attributes.rb:110:in `__update_attributes'
...
```

### DIFF
--- a/lib/tire/model/persistence/attributes.rb
+++ b/lib/tire/model/persistence/attributes.rb
@@ -116,6 +116,8 @@ module Tire
           #
           def __cast_value(name, value)
             case
+              when value == nil
+                value
 
               when klass = self.class.property_types[name.to_sym]
                 if klass.is_a?(Array) && value.is_a?(Array)


karmi:
Hi, this is a tricky one. First, I'm not sure how probable the case you describe is, and feels like a bug in the application code.

Clearly, the `@company.jobs` property expects an Array of objects (even and empty one), not `nil`, which is very toxic in this scenario. In later code, you almost certainly would like to call `@company.jobs.each ...`, and it would fail on this particular object -- not something you want to.

Also, thanks for the patch, but a solution like this has to be accompanied by a unit test, to document and define the behaviour. It is unfortunately not possible to include patches which don't include rigorous test coverage.

I've sketched a quick solution to your problem in https://github.com/karmi/tire/compare/issues/558, which I also don't feel very good about, but which is maybe closer to the intended solution...

barakcoh:
Hi, treating nil collection as empty arrays sounds good to me and more consistent with ActiveModel associations.

Will you be merging your commit into master?

P.S.
Sorry for not adding regression tests. This is my first pull request and I figured I'd add tests if you were ok with my solution.

karmi:
> Will you be merging your commit into master?

I still haven't decided what's the proper handling here... I personally think that passing `nil` to collection properties is a bug in application code (which admittedly should handled by a nice exception in Tire), and don't have enough use cases to support even my suggestion.

What other users of the persistence layer think? //cc @vhyza

vhyza:
I think there is some issue in application too. Where are you passing `nil` to collection?

Your `Company` definition returns `nil` when nothing is passed to `Company#jobs` method

``` ruby
company = Company.new name: 'foo'
p company.jobs

# nil
```

You can modify `Company` class to have `Company#jobs` method as blank array by default.

``` ruby
class Company

  property :name,         :analyzer => 'keyword'  
  property :jobs,         :class    => [Job], :default => []
end

company = Company.new name: 'foo'
p company.jobs

# []
```

So after that your `Company` instances should not return `nil`

barakcoh:
Seems to me this should be the default for all collections.

karmi:
Hi, possibly, we'll have a look into this.

I understand the sentiment 100%, but really: the solution to this is to _never_ pass `nil` to the method which expects an Array, Enumerable, etc...

Closing the issue for now...

karmi:
@barakcoh The problem should be fixed on current master.

Now, with a class definition like this:

```
class Article
  # ...
  property :comments, :class => [Comment]
end
```

the `comments` property is by default casted as an empty Array:

```
article = Article.new :title => 'Test'
article.comments
# => []
```

Do note, that you still can't pass a `nil` value to the property. So doing something like this:

```
article = Article.new :title => 'Test', :comments => nil
```

**will** break your application.

---

/cc @vhyza

barakcoh:
looks great, thx and keep up the great work..

PR #561, opened 2012-12-28
Corrects to_curl for searches containing single quotes

lemig:
corrects to_curl for searches containing single quotes

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -146,7 +146,8 @@ module Tire
       end
 
       def to_curl
-        %Q|curl -X GET '#{url}#{params.empty? ? '?' : params.to_s + '&'}pretty' -d '#{to_json}'|
+        to_json_escaped = to_json.gsub("'",'\u0027')
+        %Q|curl -X GET '#{url}#{params.empty? ? '?' : params.to_s + '&'}pretty' -d '#{to_json_escaped}'|
       end
 
       def to_hash


karmi:
Hi, good catch! What about replacing the single quotes with escaped format (`\'`) instead?

lemig:
The escaped ( \' ) format closes the quotes around the request body.

karmi:
So, what's an example of before & after format?

lemig:
Thanks for your prompt review.

According to http://bit.ly/UsESzY, single quotes do not need to be escaped in Lucene.
So this is OK:

```
s = Tire.search 'articles' do
  query do
    string "l'amour"
  end
end
```

But:

```
puts s.to_curl
```

Produces invalid:

```
curl -X GET 'http://localhost:9200/articles/_search?pretty' -d '{
  "query":{
    "query_string":{
      "query":"l'amour"
    }
  }}'
```

After format:

```
curl -X GET 'http://localhost:9200/articles/_search?pretty' -d '{
  "query":{
    "query_string":{
      "query":"l\u0027amour"
    }
  }}'
```

Also invalid ( \' ):

```
curl -X GET 'http://localhost:9200/articles/_search?pretty' -d '{
  "query":{
    "query_string":{
      "query":"l\'amour"
    }
  }}'
```

Other example: http://bit.ly/12NUNfZ

karmi:
Thanks!, merged.

PR #567, opened 2013-01-05
Add: Definition to check document exists?

palanglung:
I think in some cases we need to check document existence without retrieving any document body,
in my case i use random ordered log for indexing sometime the main log object logged before the child object so we need to check the document existence before indexing. if document fail the log will go back to queue and if document exist will update document with the child log.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -276,7 +276,7 @@ module Tire
     def update(type, id, payload={}, options={})
       raise ArgumentError, "Please pass a document type" unless type
       raise ArgumentError, "Please pass a document ID"   unless id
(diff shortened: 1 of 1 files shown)


karmi:
> in my case i use random ordered log for indexing sometime the main (...)

I'm not sure I understand your use case, but here are some notes regarding the patch:
- Please remove the redundant commits, use `git rebase` to get rid of merge commits
- It is unfortunately not possibly to accept patches without unit tests (and integration tests, where warranted)
- I'm concerned about naming -- `document_exists?` does not sound right. Maybe `has_document?`? Try come up with a good name.
  I'm worried that if we use this naming scheme, then `Index#store` should be named `Index#store_document` etc. Any thoughts?

PR #571, opened 2013-01-07
Added support for replication=async option when storing index document

kailuowang:
It's a very simple change to support Asynchronous Replication option when indexing a document on elasticsearch

-----------From Elastic Search Website---------
Asynchronous Replication
By default, the index operation only returns after all shards within the replication group have indexed the document (sync replication). To enable asynchronous replication, causing the replication process to take place in the background, set the replication parameter to async. When asynchronous replication is used, the index operation will return as soon as the operation succeeds on the primary shard.

Thanks

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -79,6 +79,7 @@ module Tire
 
       params[:parent]  = options[:parent]  if options[:parent]
       params[:routing] = options[:routing] if options[:routing]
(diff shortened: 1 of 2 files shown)


karmi:
Hi, sorry, it took a bit longer to merge. Just a word of caution: know what you're doing when you would be using the asynchronous replication. Elasticsearch is plenty fast with the default replication, any bottleneck you feel might be elsewhere.

PR #572, opened 2013-01-07
Rename @attributes to @attrs

matthewford:
I was getting a name clash with AR @attributes so I have renamed it


karmi:
Hi, I've added integration tests for the percolation support in models, retouched and merged your patch. Thanks!

PR #581, opened 2013-01-14
Fixes bug in to_indexed_json example

bloudermilk:
The previous example attempted to disable JSON roots but didn't reference `self` so it only assigned a variable.


karmi:
Hi, thanks!, sorry it took so long.

PR #583, opened 2013-01-14
Casting times formatted with either a "Z" or a time zone offset

chrislo:
The casting of strings to Time in persistence/attributes.rb assumed strings with a trailing "Z" (i.e. in UTC time). This pull request adds support for strings with a time zone offset (e.g. +01:00).


karmi:
Thanks!, that makes sense. I also added that the time is _forced_ into UTC after parsing.

karmi:
Closed by 69fa54b061884143f55e6c3c50e4ff4fe23fb2ca

PR #588, opened 2013-01-16
simplify rake test by combine unit and integration

timoschilling:


karmi:
But they don't do the same thing, AFAIK...

PR #589, opened 2013-01-16
Allow Tire to infer the wrapper class

mrcasals:
Set `Tire.configuration.wrapper` to `:infer` and Tire will infer the wrapper
class for search results.

This is useful when your search query returns results with different classes.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -357,6 +357,11 @@ If that seems like a great idea to you, there's a big chance you already have su
 
 One would bet it's an `ActiveRecord` or `ActiveModel` class, containing model of your Rails application.
 
+If your search results return items with different classes, you can set
+`Tire.configuration.wrapper` to `:infer` and Tire will try to use the class
+from the item `_type` property (i.e. if `_type` is `article` the inferred
+class will be `Article`).
+
 Fortunately, _Tire_ makes blending _Elasticsearch_ features into your models trivially possible.
 
 
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -57,14 +57,6 @@ module Tire
       end
       alias :[] :slice
 
-      def to_ary
-        self
-      end
-
-      def as_json(options=nil)
-        to_a.map { |item| item.as_json(options) }
-      end
-
       def error
         @response['error']
       end
@@ -75,6 +67,10 @@ module Tire
 
       def failure?
         ! success?
+      end
+
+      def to_ary
+        self
       end
 
       # Handles _source prefixed fields properly: strips the prefix and converts fields to nested Hashes
@@ -113,8 +109,22 @@ module Tire
             # Update the document with meta information
             ['_score', '_type', '_index', '_version', 'sort', 'highlight', '_explanation'].each { |key| document.update( {key => h[key]} || {} ) }
 
-            # Return an instance of the "wrapper" class
-            @wrapper.new(document)
+            if @wrapper == :infer
+              begin
+                type = document['_type']
+                klass = type.camelize.constantize
+
+                # Return an instance of the _type class
+                klass.new(document)
+              rescue NameError => e
+                raise NameError, "You have tried to infer the results wrapper, but " +
+                                 "Tire cannot find the model class '#{type.camelize}' " +
+                                 "based on _type '#{type}'.", e.backtrace
+              end
+            else
+              # Return an instance of the "wrapper" class
+              @wrapper.new(document)
+            end
           end
         end
       end
--- a/test/unit/results_collection_test.rb
+++ b/test/unit/results_collection_test.rb
@@ -149,6 +149,14 @@ module Tire
           assert_equal   'Test',  article.title
         end
 
+        should "allow wrapping hits in inferred class" do
+          Configuration.wrapper(:infer)
+
+          article =  Results::Collection.new(@response).first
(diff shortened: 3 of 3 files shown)


mrcasals:
Ping

mrcasals:
Ping. What do you think about this, @karmi?

karmi:
> If your search results return items with different classes (...)

@mrcasals I'm sorry for the silence, and also, I'm afraid I don't understand the intention here...

karmi:
@mrcasals Reviewing it again -- the problem with this is that the corresponding part of code is complicated and "magical" quite a lot already. Why not simply `map` the results in application code before you use them?

``` ruby
Tire.search.results.map do |result|
  klass = document['_type'].camelize.constantize
  klass.new result.to_hash
end
```

mrcasals:
@karmi sorry for the delay on the answer. The problem I found is that when my search returns results and all of them belong to the same class, Tire correctly wraps the results with the model class (so, if my search returns only products, Tire will wrap them with my `Product` class), so I will be able to use my own defined methods.

But if my search returns elements from two different classes (let's say, `Product` and `Business`), Tire will not wrap the results and will use its own `Collection` method. I added the `:infer` wrapper option that will automatically wrap the results with their own class. Hope this helps to understand the issue.

Also yeah, I can totally do what you say, but as there is a case when Tire will automatically wrap the results with my method, having to wrap them myself is unexpected to me :)

karmi:
I understand the general intention, thanks. The thing is, Tire is already quite sophisticated when you're eg. using search and models in a Rails application. Notice:

```
>> s = Tire.search( ['authors','articles'] ) { query { string 'title:"Title 3" OR name:*' }  }

>> s.results[0]
=> <Item (Article) author: nil, content: "Lorem", id: "12716", published_on: "2013-01-02", title: "Title 3", _score: 211.04463, _type: "article", _index: "articles", _version: nil, sort: nil, highlight: nil, _explanation: nil>

>> s.results[0].class
=> Article(id: integer, title: string, content: text, author: string, published_on: date, created_at: datetime, updated_at: datetime)

>> s.results[1]
=> <Item (Author) birthday: nil, created_at: "2013-03-07T08:13:25Z", name: "Franz Kafka", updated_at: "2013-03-07T08:13:25Z", id: "1", _score: 0.2236068, _type: "author", _index: "authors", _version: nil, sort: nil, highlight: nil, _explanation: nil>

>> s.results[1].class
=> Author(id: integer, name: string, birthday: date, created_at: datetime, updated_at: datetime)

>> s = Tire.search( ['authors','articles'], load:true ) { query { string 'title:"Title 3" OR name:*' }  }

>> s.results[0]
  Article Load (0.4ms)  SELECT "articles".* FROM "articles" WHERE "articles"."id" = ? LIMIT 1  [["id", "12716"]]
  Author Load (0.2ms)  SELECT "authors".* FROM "authors" WHERE "authors"."id" = ? LIMIT 1  [["id", "1"]]
=> #<Article id: 12716, title: "Title 3", content: "Lorem", author: nil, published_on: "2013-01-02", created_at: "2013-01-04 19:34:00", updated_at: "2013-01-04 19:34:00">

>> s.results[1]

=> #<Author id: 1, name: "Franz Kafka", birthday: nil, created_at: "2013-03-07 08:13:25", updated_at: "2013-03-07 08:13:25">
>> s.results[1].class
=> Author(id: integer, name: string, birthday: date, created_at: datetime, updated_at: datetime)
```

So, it already “infers” the class from the hit `_type`.

I guess I'd need more context here, specifically:
- Is this within Rails or not
- Is this an ActiveModel/Record integration?
- Is it Tire::Model::Persistence?
- Is this in a generic Ruby, model-less application, using straight Ruby classes?

Can you expand on this, please?

mrcasals:
It's a Rails app, but I'm using Tire's `Tire::Model::Persistence` to get an ActiveModel/Record API. Also, I'm using the imperative way to work with Tire instead of its DSL.

karmi:
>  I'm using Tire's `Tire::Model::Persistence` to get an ActiveModel/Record API. 

Now I'm even more confused .) `Tire::Model::Persistence` or `Tire::Model::Search`? In Rails, the result item should realy be either "masqueraded" as your models (without `:load`), or instances of your models (with `load`). Can you try the code I posted and post what's not working in your environment? (Or, preferably, add a failing integration test.)

mrcasals:
Yes, your code works as expected.

My app has two models, and they include `Tire::Model::Persistence` to `save` and `find` objects. I need to search results from both of my models in the same search, so I use the `Tire::Search::Search` class and [URI Parameters](http://www.elasticsearch.org/guide/reference/api/search/uri-request.html) (more precisely, I'm using the `q`, `sort`, `size`, `type` parameters).

I'm having problems understanding Tire's codebase (too much magic involved, I guess) and I don't find the DSL comfortable to work with, that's why I started using the URI Parameters. 

Anyway, my point was: using `Tire::Search::Search`, if the search returns results from 2 different models they are not properly wrapped. That's why I opened this PR :)

karmi:
Good!, but the problem is until I can replicate the issue with a failing test or a code snippet, it's hard to decide what should be done :) The Tire DSL is just a wrapper around `Tire::Search::Search` etc classes...

I'll close the issue for now, and ask you to reopen it with an executable code snippet / integration test for further discussion, does it sound fair?

PR #594, opened 2013-01-20
Safe response body parsing

andrewslotin:
Sometimes the ElasticSearch server returns an empty body response with code 404. Parsing an empty string makes MultiJson throwing an exception, which is not expected result in that case.
With following changes Tire::Index will try to parse response body only on success. Otherwise `nil` will be returned.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -440,7 +440,12 @@ module Tire
     protected
 
     def parse_response(response)
-      MultiJson.decode(response.body) if response.success?
+      if response.success?
+        MultiJson.decode(response.body)
+      else
+        Configuration.logger.log_response(response.code, nil, response.body) if Configuration.logger
+        nil
+      end
     end
 
   end
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -54,12 +54,12 @@ module Tire
 
     def mapping
       @response = Configuration.client.get("#{url}/_mapping")
-      MultiJson.decode(@response.body)[@name]
+      (parse_response(@response) || {})[@name]
     end
 
     def settings
       @response = Configuration.client.get("#{url}/_settings")
-      MultiJson.decode(@response.body)[@name]['settings']
+      (parse_response(@response) || {})[@name]['settings']
     end
 
     def store(*args)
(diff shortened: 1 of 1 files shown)


karmi:
It should return empty response only for HEAD requests?

andrewslotin:
@karmi, the main goal is to check if the request was successful (i.e. server responded with 200 etc.) Sometimes ElasticSearch returns an empty response (i.e. when the search server is gone).

karmi:
>  i.e. when the search server is gone

Can you describe it in more detail, please?

andrewslotin:
I'm putting my ElastcSearch server down and trying to call `Tire::Index#retrieve` which causes multi_json to raise `MultiJson::DecodeError` with an error message saying that I'm trying to parse an empty string as JSON, which is not valid.

karmi:
I'm sorry, but I can't replicate it:

```
>> Tire.index('articles').retrieve(:article, 1)
Errno::ECONNREFUSED: Connection refused - connect(2)
```

I'll close the issue for now, please reopen with any executable code I can run to demonstrate the issue!

andrewslotin:
Ok, got it. Looks like this issue can be reproduced when using Faraday as an HTTP client. Here is a code that replicates the issue:

```
require 'typhoeus'
require 'tire/http/clients/faraday'

Tire.configure do |config|
  Tire::HTTP::Client::Faraday.faraday_middleware = Proc.new do |builder|
    builder.adapter :typhoeus
  end
  config.client Tire::HTTP::Client::Faraday
end

Tire.index('articles').retrieve(:article, 1)
```

Anyway, checking if request was successful seems to be good idea :smile: 

(Cannot reopen this issue, should I make another pull request?)

karmi:
> (...) checking if request was successful (...)

But most probably not in `Tire::Index`. The problem does not exist with the default client (`Errno::ECONNREFUSED`) or the Curb client (`Curl::Err::ConnectionFailedError`). This is something which should be fixed in the client itself, not adding `if...else` routines to a completely different class?

andrewslotin:
@karmi, sounds fair. But this makes the gem to be dependent on implementation of HTTP client, right?

karmi:
In what sense? I didn't wrote the Faraday adapter -- there should be a unit test which makes sure it behaves the same as RestClient/Curb clients. What we're seeing here is the Faraday adapter returning response even when the Elasticsearch service is down -- this hints at the adapter is behaving incorrectly.

PR #601, opened 2013-01-24
Fix spelling errors in the README.

ckrailo:
Mostly so I could find both places for "associations".

(Sorry for such a silly pull request. Maybe somebody else will search for "associations" too and will benefit.)


karmi:
Sorry for the delay!, merged/pushed.

PR #602, opened 2013-01-25
Load includes with mongoid

### LINKED ISSUE #282: eager load and mongoid
Hi,

I just want to use the [mongoid eager loading](http://mongoid.org/docs/querying/criteria.html#includes) (includes) with tire.
But It does not work (subjects is the name of my relation)

```
#query = Product.includes(:subjects).tire.search :load => true do |search|
query = Product.tire.search :load => {:includes => :subjects} do |search|
...
end
p query.results
```

```
def load(options=nil)
      options ? self.class.find(self.id, options) : self.class.find(self.id)
end
```

How can I do to use both earger loading (thin and mongoid) ?

Thanks

timoschilling:
before this PR the loading will be done by:

``` ruby
klass.find(ids, @options[:load])
```

but this doesn't work with Mongoid, Mongoid need's this way:

``` ruby
klass.find(ids).includes(@options[:load][:include])
```

This also closes #282

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -148,7 +148,15 @@ module Tire
       end
 
       def __find_records_by_ids(klass, ids)
-        @options[:load] === true ? klass.find(ids) : klass.find(ids, @options[:load])
+        if @options[:load] === true
+          klass.find(ids)
+        else
+          if @options[:load][:include] && defined?(Mongoid) && klass < Mongoid::Document
+            klass.includes(@options[:load][:include]).find(ids)
+          else
+            klass.find(ids, @options[:load])
+          end
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


karmi:
@timoschilling Hmm, thanks for looking into this. I don't like conditional logic based on `defined?(Something)`, the code in question is messy already. I think we need to provide a way to `require "something/special"`, or perform this in a specific module. When the methods do not provide enough hooks for that, we need to refactor them so they do.

Notice that the `__find_records_by_ids` method was added _exactly_ for purposes like these -- but it should be overloaded by a `Something` module, not changed. What do you think?

timoschilling:
The `defined?(Mongoid)` check is not the 'reale' condition, it is only to make `klass < Mongoid::Document` save.

I think the `require` way is not good!
1. Tire with Mongoid works out of the Box, why i need to require something for eager loading of Mongoid?
2. Overwriting of `__find_records_by_ids` is not good while whats happen if you want to load Mongoid and ActiveRecord at one time?

karmi:
@timoschilling Hi, I think the proper way around this would be to implement a `Strategy` for each specific ORM/persistence library, like we did in the `Mymodel#import` code in efb9c14. Then I'm all for having it included as part of the core gem. I'll try to have a look into that soon.

timoschilling:
@karmi Strategies is a good way! Give me 1 week, than i implement it.

PR #603, opened 2013-01-28
Update README.markdown

kroepke:
Mention that for ActiveRecord one should be using after_commit and after_rollback.


kroepke:
well 'ree' doesn't work…

Issue #582, opened 2013-01-14
Dynamic attributes for persistent modes


davekinkead:
@karmi, Tire + Elasticsearch is a fantasticly simple nosql persistence option for model data.

One thing I love about nosql is the schema-free nature of it meaning rapid prototyping is a breeze.  I often work with evolving data schema but the flexibility of ES is limited to some extend by the declarative requirements of Tire - requiring the property method for any attr that I want to access.

So is there an easy way of making the following work out of the box? Ideally I shouldn't have to change my model code if I need to add additional attribute on the fly.

```
class Foo
  include Tire::Model::Persistence
end

foo = Foo.create :name => 'foo', :bar => 'baz'
foo.bar # => I want 'baz' but currently can't access it
```

davekinkead:
Looking under the hood a bit more, it seems the problem can be addressed with the following:

```
class Foo

  include Tire::Model::Persistence

  # => Override constructor to implement Tire Persistence
  def initialize(attrs={})
    attrs.each do |attr, value|
      # => call Tire's property method
      self.class.property attr unless self.class.property_types.keys.include? attr
      # => set instance variable
      instance_variable_set("@#{attr}", value)
    end
  end

end
```

Any thoughts to making this default behaviour?

karmi:
Hi,

thanks for your thoughts and your code! I understand the issue, and understand why it would help when prototyping or working with heterogenous data.

The problem is, that a “sugar” like this quickly brings hypoglycemia on when you're not paying enough attention :) Models are initialized with heterogenous properties, those are saved into Elasticsearch with dynamic mapping, then somebody wants to sort on the `name` property and it does not work...

I do agree a default like this would be aligned with Elasticsearch's behaviour, but I'm just worried about the "sugar levels" here too much, based on experience.

That said, I think it would make a terrific [`tire-contrib`](https://github.com/karmi/tire-contrib) module, where you could use it like this:

``` ruby
    require 'tire'
    require 'tire/model/persistence/dynamic_attributes'

    class Foo
      include Tire::Model::Persistence
      include Tire::Model::Persistence::DynamicAttributes
    end
```

What do you think?

davekinkead:
I have to admit that I haven't ventured into the tire-contrib module as the docs are very sparse. 

If you think that naming clashes could be a problem when searching, then the contrib module is fine.  I'm currently extending a base model class so this would do that in one _include_ line.

[...]

PR #605, opened 2013-01-29
Permit dynamic declaration of persistent properties

davekinkead:
From issue #582. 

Having to explicitly declare attributes with the property method sours the sweetness of Elasticsearch's schema-less goodness.

This pull request allows attributes to be declared dynamically at instantiation without affecting explicit property mappings.

### DIFF
--- a/lib/tire/model/persistence/attributes.rb
+++ b/lib/tire/model/persistence/attributes.rb
@@ -87,6 +87,14 @@ module Tire
           attr_accessor :id
 
           def initialize(attributes={})
+  
+            attributes.each do |attr, value|
+              # call Tire's property class method
+              self.class.property attr unless self.class.property_types.keys.include? attr
+              # set instance variable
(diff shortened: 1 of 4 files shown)


karmi:
Hi, as suggested in #582, do you think it would make sense as a  [`tire-contrib`](https://github.com/karmi/tire-contrib) module?, where the end-user could use it somehow like this:

``` ruby
    require 'tire'
    require 'tire/model/persistence/dynamic_attributes'

    class Foo
      include Tire::Model::Persistence::DynamicAttributes
    end
```

(Note to self: I really need to write more about tire-contrib and less invasive ways of extending Tire :)

karmi:
@davekinkead Any thoughts about doing it as Tire contrib extension?

davekinkead:
@karmi happy to do this as a contrib - just need to find some time this week :(

karmi:
No worries, Dave! Take your time and thanks!

PR #617, opened 2013-02-05
Callback Improvement

AaronRustad:
Currently, Model::Callbacks updates the index on any change to the model even if the attributes being changed are not part of the index being updated. This PR includes changes that allow the user to specify if the Model should update ElasticSearch on indexed-attribute changes only.

In the below example, only changes to the :name attribute will cause the ElasticSearch server to be updated.

```
class SomeModel < ActiveRecord
  should_tire_only_update_on_index_changes true
  tire.mapping do
    indexes :name, analyzer: 'snowball'
  end
end    
```

### DIFF
--- a/lib/tire/model/callbacks.rb
+++ b/lib/tire/model/callbacks.rb
@@ -1,3 +1,4 @@
+require 'set'
 module Tire
   module Model
 
@@ -11,6 +12,10 @@ module Tire
     #
     module Callbacks
 
+			def self.save_only_on_indexed
+				true
+			end
+
       # A hook triggered by the `include Tire::Model::Callbacks` statement in the model.
       #
       def self.included(base)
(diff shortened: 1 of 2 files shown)


karmi:
> If Developers want this level of custom callback behavior, it's possible they should be writing their own callbacks.

@darrenboyd I couldn't agree _more_ :) I have been always advising people to use the `include Tire::Model::Callbacks` as the initial approach, and see how far does it carry them. It is very easy to use custom logic -- see the `update_index if state == 'published'` part in the README.

karmi:
@AaronRustad I understand and even support the motivation for this! That said, let me briefly describe what I don't like about this approach:
- The usage of `really_weird_acts_as_something_or_maybe_not` class methods is a relic of the Rails's past, and we should actively stay away from it :)
- The ActiveModel integration itself should bring ActiveModel::Dirty into account, which, shamefully, it does not. In your case, you only include it for one model.
- As said above, I'm not a fan of super-smart default behaviour. As a library developer, I want to a) make it easy for people to use the technology, b) provide a rich API to adapt and enhance the default behaviour. In this case, it is very easy to write a custom `after_save`/`after_commit` hook which checks for dirty attributed, `published=true` conditions, etc. I think it's a much more maintanable way of achieving your goal then relying on smart defaults in thr library.

AaronRustad:
Fair enough. Thanks @darrenboyd and @karmi for your constructive criticism.

karmi:
@AaronRustad OK, should we close this issue then?

PR #641, opened 2013-02-25
Ignore :payload options for Search#params.

metaskills:
When building a Tire.search DSL with a hash of options for the payload, we need to ignore the `:payload` in `params` so the `to_curl` and logging does not include a bunch of noise.

You may want to consider ignoring :payload in these two places too. Thanks!
- https://github.com/karmi/tire/blob/master/lib/tire/count.rb#L26
- https://github.com/karmi/tire/blob/master/lib/tire/multi_search.rb#L181

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -48,7 +48,7 @@ module Tire
       end
 
       def params
-        options = @options.except(:wrapper)
+        options = @options.except(:wrapper, :payload)
         options.empty? ? '' : '?' + options.to_param
       end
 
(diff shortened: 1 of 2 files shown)


karmi:
That makes sense, thanks for looking into it. I'll try to get to this soon.

metaskills:
Thanks! BTW, I had two test failures on my first run. But they went away when I ran them the second time.

karmi:
Yes, I know, they are related to percolator and dis max I think, didn't yet found a good solution for them.

PR #643, opened 2013-02-26
Allow support of boost (and other options) to Terms Query

moktin:
Hi,

While using tire, I had problem using the boost on query terms.

I am pretty new to both tire and elasticsearch world, so maybe I am just doing it wrong.

With this indices

``` ruby
class User < ActiveRecord::Base
  include Tire::Model::Search
  mapping do
    indexes :comments do
      indexes :id, type: :integer
    end
  end
end
```

A request like : 

``` ruby
search = User.search do
  query do
    boolean do
      should { 
        terms 'comments.id',  [1, 2], boost: 10
      }
    end
  end
end
```

produce such a result : 

```
curl -X GET 'http://localhost:9200/users/user/_search?load=true&size=12&pretty' -d '
  {
    "query": {
      "bool": {
        "should": [
          {
            "terms": {
              "comments.id": [
                1,2
              ]
            }
          }
        ]
      }
    },
    "size": 12
  }
'
```

When I was waiting for something like that : 

```
curl -X GET 'http://localhost:9200/users/user/_search?load=true&size=12&pretty' -d '
  {
    "query": {
      "bool": {
        "should": [
          {
            "terms": {
              "comments.id": [
                1,2
              ],
              "boost": 10
            }
          }
        ]
      }
    },
    "size": 12
  }
'
```

The attached pull request solve that.

Thank you to let me know if it helps or if you need some more informations

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -20,7 +20,7 @@ module Tire
 
       def terms(field, value, options={})
         @value = { :terms => { field => value } }
-        @value[:terms].update( { :minimum_match => options[:minimum_match] } ) if options[:minimum_match]
+        @value[:terms].update(options)# if options.present?
         @value
       end
 
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -20,7 +20,7 @@ module Tire
 
       def terms(field, value, options={})
         @value = { :terms => { field => value } }
-        @value[:terms].update(options)# if options.present?
+        @value[:terms].update(options)
         @value
       end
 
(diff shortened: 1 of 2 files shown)


brupm:
:+1: Very useful PR. Although the build is failing.

moktin:
Hi,
The test are green locally :  

```
Finished in 173.233226 seconds.

774 tests, 774 passed, 0 failures, 0 errors, 0 skips, 1745 assertions
```

I also noticed that most of the past builds on Travis are failing (see https://travis-ci.org/karmi/tire/builds) and it seems to be exactly the same errors.

@karmi are the builds having general problem on travis ? Or are they "real" failure ?

karmi:
Thanks!, merged & pushed.

---

Re the Travis failures -- there has been some incompatibilites with Ruby 1.8, which were shooting down the tests, and the percolator tests are still failing consistently. I have tried to fix it in 1a8b921, but will need to have a better look into it.

Issue #530, opened 2012-11-25
Bulk import MYSQL table using Tire


visasquare:
Hi Karmi, 

Is there a way that I can bulk import my documents(part of mysql table) into elastic search. Till now I have been using river-jdbc but I really want to do it using tire as I get lot of flexibility in defining multi field indexes using Tire.

Thanks!

visasquare:
Got the answer by reading through the documents. we can set the pagination and acheive bulk import

:per_page => 1000

Thanks Karmi!

karmi:
Hi, yes, you can pass specific `:per_page` on the command line:

```
rake environment tire:import CLASS='Article' INDEX='articles-2011-05' PARAMS={per_page:10000}
```

When you want to index a specific subset of records, with a specific SQL query, it may well pay off to write your own, custom indexing code. [`ActiveRecord#find_in_batches`](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches) makes it very convenient:

``` ruby
Article.where("published_on > ?", Time.parse("2012-10-01")).find_in_batches do |group|
  Tire.index('myindex').import group
end
```

PR #652, opened 2013-03-01
Improve the `tire:import` Rake task

jwaldrip:
I wanted to submit this as a topic for conversation. This solution provides the following:
- Detection of models using Tire::Model::Search.mapping, auto-indexes all of these by default.
- You can still pass CLASS var, but it now accepts multiples.
- Removed Kaminari, not sure why it was needed in this task.
- Uses the progress_bar gem, a very clean implementation IMHO.

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -1,3 +1,5 @@
+require 'set'
+
 module Tire
   module Model
 
@@ -18,6 +20,9 @@ module Tire
     #
     #
     module Search
+
+      mattr_accessor :dependents
+      self.dependents = Set.new
 
       # Alias for Tire::Model::Naming::ClassMethods.index_prefix
       #
@@ -254,6 +259,7 @@ module Tire
       # A hook triggered by the `include Tire::Model::Search` statement in the model.
       #
       def self.included(base)
+        self.dependents << base
         base.class_eval do
 
           # Returns proxy to the _Tire's_ class methods.
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -121,10 +121,10 @@ namespace :tire do
     full_comment_drop = <<-DESC.gsub(/      /, '')
       Delete indices passed in the INDEX environment variable; separate multiple indices by comma.
 
-      Pass name of a single index to drop in the INDEX environmnet variable:
+      Pass name of a single index to drop in the INDEX environment variable:
         $ rake environment tire:index:drop INDEX=articles
 
-      Pass names of multiple indices to drop in the INDEX or INDICES environmnet variable:
+      Pass names of multiple indices to drop in the INDEX or INDICES environment variable:
         $ rake environment tire:index:drop INDICES=articles-2011-01,articles-2011-02
 
     DESC
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -72,7 +72,8 @@ namespace :tire do
       # Create the index if it doesn't exist
       unless index.exists?
         mapping = MultiJson.encode(klass.tire.mapping_to_hash, :pretty => Tire::Configuration.pretty)
-        puts "[IMPORT] Creating index '#{index.name}' with mapping:", mapping
+        puts "[IMPORT] Creating index '#{index.name}' with mapping:",
+             mapping
         unless index.create(:mappings => klass.tire.mapping_to_hash, :settings => klass.tire.settings)
           STDERR.puts "[ERROR] There has been an error when creating the index -- elasticsearch returned:",
                       index.response
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -81,25 +81,14 @@ namespace :tire do
         end
       end
 
-      puts "[IMPORT] Starting import for the '#{klass}' class"
-      progress_bar = ProgressBar.new(klass.count)
-
-      # Use an indexer scope if it is defined
-      klass = klass.indexer if klass.respond_to? :indexer
-
-      # Try and use AR find_in_batches
-      if klass.respond_to?(:find_in_batches)
-        klass.find_in_batches do |group|
-          index.import(group, params) do |documents|
-            GC.start
-            progress_bar.increment! documents.count
-            documents
+      # Add Pagination to the class if it doesn't exist
+      if defined?(Kaminari) && klass.respond_to?(:page)
+        klass.instance_eval do
+          def paginate(options = {})
+            page(options[:page]).per(options[:per_page]).to_a
           end
         end
-      elsif klass.respond_to? :all
-        index.import(klass.all, params) do |documents|
-          GC.start
-          progress_bar.increment! documents.count
+      end unless klass.respond_to?(:paginate)
           documents
         end
       else
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -1,105 +1,115 @@
 require 'rake'
-require 'benchmark'
+require 'progress_bar'
+require 'set'
 
 namespace :tire do
 
   full_comment_import = <<-DESC.gsub(/    /, '')
     Import data from your model using paginate: rake environment tire:import CLASS='MyModel'.
 
-    Pass params for the `paginate` method:
-      $ rake environment tire:import CLASS='Article' PARAMS='{:page => 1}'
+    import all models:
+      $ rake tire:import
 
     Force rebuilding the index (delete and create):
-      $ rake environment tire:import CLASS='Article' PARAMS='{:page => 1}' FORCE=1
+      $ rake tire:import FORCE=1
 
-    Set target index name:
-      $ rake environment tire:import CLASS='Article' INDEX='articles-new'
+    Set target classes:
+      $ rake tire:import CLASS='Article, Comment'
   DESC
   desc full_comment_import
-  task :import do |t|
-
-    def elapsed_to_human(elapsed)
-      hour = 60*60
-      day  = hour*24
-
-      case elapsed
-      when 0..59
-        "#{sprintf("%1.2f", elapsed)} seconds"
-      when 60..hour-1
-        "#{(elapsed/60).floor} minutes and #{(elapsed % 60).floor} seconds"
-      when hour..day
-        "#{(elapsed/hour).floor} hours and #{(elapsed/60 % hour).floor} minutes"
-      else
-        "#{(elapsed/hour).round} hours"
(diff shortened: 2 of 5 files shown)


gerred:
:space_invader: :+1:

karmi:
Hi Jason, thanks for the pull request, it shows that you care a lot about the library, and I'm very happy to see that. There are, however, many issues with the proposed changes.

The first -- and most important -- one is that the Rake task is a convenience, a simple way how to get started and import your existing data into Elasticsearch. There are many potential pitfalls -- for instance, when importing large amounts of data from a RDBMS, you'll probably want to use `ActiveRecord::Base#find_in_batches` instead of pagination [#530].

I think we should _encourage_ people to write their own importing code, instead of trying to support every corner case in the Rake task, and make it unnecessary smart. This is Ruby, after all, and Tire gives you a rich API for custom importing routines and workflows.

Now, there are some practical issues with the proposed changes as well:
- I don't like the dependency on the `progress_bar` gem. Any dependency is a liability and I'd like to keep it down as much as possible. I agree the custom progress bar code could be replaced, but I'd use eg. [`ANSI::ProgressBar`](http://rubydoc.info/github/rubyworks/ansi/master/ANSI/ProgressBar), which is part of the `ansi` gem. (The Turn gem already brings Ansi as a developlment dependency.) I just don't think it's worth it, since `ansi` gem is widely used with wildly varying version constraints and you have to `bundle exec` everything.
- Automatically depending on `environment` is precisely the sort of pattern I don't like in gems. What if I'm _not_ inside a Rails application, what if I'm using Sinatra or there just isn't the `environment` task? What if I want to write my own Rake task and just `invoke` the Tire one?
- I'm not sure about the utility of importing multiple models in one go. Again, I understand how it might simplify certain use cases, but notice how you need to manipulate `TOTAL_COUNT`, split the environment variable... Why, instead of adding complexity to the Tire task, a person who needs this, doesn't simply write a custom task which just invokes the Tire task multiple times?
- Any "automatic detection" of anything is a nightmare for me :) First thing, it has to work, which I bet it does -- but it's very hard to test in automated manner. So for any refactoring/change of this code, you have to have a Rails app ready, and try it manually. Believe me, it takes a _lot_ of my time to keep checks on things like that, making sure the task works for ActiveRecord as well as for Mongoid, etc. I'm a fan of "prefer declaration over introspection" in a situation like this.
- I love the introspection on `find_in_batches`, though! I think the task isn't a good place for it, since it is not part of the Tire Ruby API. [`Model::Import::ClassMethods#import`](http://rubydoc.info/gems/tire/Tire/Model/Import/ClassMethods:import) and [`Import#index`](http://rubydoc.info/gems/tire/Tire/Index:import) would be a _much_ better place, and we already support `method`. The model could do the introspection and pick either `paginate` or `find_in_batches` -- or any other "bulk find" method supported by OxM libraries. We should then skip [defaulting](https://github.com/karmi/tire/blob/master/lib/tire/tasks.rb#L45) to `paginate` in the task.
- The `paginate` method has ben injected in the task as a convenience in 3925c24, since with _Kaminari_, there is no `paginate` method by default -- this forced users to write their own `paginate` method just for the import task. See issue #192.
- The `klass.respond_to? :all` clause is particularly dangerous, because that would just blow up on huge datasets, without people being aware what's going on. It would just "not work", and these (rightly) frustrated users would then submit tickets. We don't want that :)
- I love the notion if `MyModel.indexer` [scope](https://github.com/jwaldrip/tire/blob/8ce911c9517ec3279b3773d5c99658c12e7ddc3e/lib/tire/tasks.rb#L84), that's very clever. It nears the thin line where people should write their own custom indexing routine, but is maybe worth it.
- Why the call to [`GC.start`](https://github.com/jwaldrip/tire/blob/8ce911c9517ec3279b3773d5c99658c12e7ddc3e/lib/tire/tasks.rb#L90) for each batch?

I'm sorry for such a long and potentially unpleasant review, but in the end, I'm the one who's there when stuff breaks or behaves weird, and need to exercise caution... I think we should focus on the great ideas you had in the patch:
- [ ] In `Model::Import::ClassMethods#import`, introspect the host class, and use `find_in_batches` if present
- [ ] Explore the notion of `MyModel.indexer` to define the scope for ActiveRecord etc. This could be part of the model DSL such as `mapping`, etc.

jwaldrip:
"Why the call to GC.start for each batch?"

The reason for this is that while importing a large set of data some odd 5million records of data with the find_in_batches call I found my memory climbing to over 3.5 gigs, GC.start fixes this in all instances.

jwaldrip:
In regards to the indexer scope, what would you think about doing the following:

``` ruby
class MyModel
  include Tire::Model::Search
  tire do
    scope do
      includes(:comments)
    end
    mapping do
       indexes :name
       indexes :comments_count, as: "comments.count"
    end
  end
end
```

lib/tire/tasks.rb
```diff
@@ -1,105 +1,115 @@
...
-      mapping = MultiJson.encode(klass.tire.mapping_to_hash, :pretty => Tire::Configuration.pretty)
-      puts "[IMPORT] Creating index '#{index.name}' with mapping:", mapping
-      unless index.create( :mappings => klass.tire.mapping_to_hash, :settings => klass.tire.settings )
-        STDERR.puts "[ERROR] There has been an error when creating the index -- elasticsearch returned:",
-                    index.response
-        exit(1)
+      # Create the index if it doesn't exist
+      unless index.exists?
+        mapping = MultiJson.encode(klass.tire.mapping_to_hash, :pretty => Tire::Configuration.pretty)
+        puts "[IMPORT] Creating index '#{index.name}' with mapping:", mapping
```
nz: nitpick: might as well interpolate `mapping` for consistency. not sure how common it is to pass multiple args to `puts` in Rubyland anyway. also line 74, though that line at least benefits from some line wrapping.

lib/tire/tasks.rb
```diff
@@ -1,105 +1,115 @@
...
         end
-      end unless klass.respond_to?(:paginate)
-
-      # Import the documents
-      #
-      index.import(klass, params) do |documents|
-
-        if total
-          done += documents.to_a.size
-          # I CAN HAZ PROGREZ BAR LIEK HOMEBRU!
```
nz, line 90: this comment definitely needed to go

lib/tire/tasks.rb
jwaldrip: Although I didn't do the original message here, I think the reason is to achieve a newline, I will update with some better wrapping.

nz:
reindexing rake tasks are fun. there are a lot of ideas in here that I like and will probably borrow for Sunspot.
- :+1: for multi-model CLASS — even better to default to all classes a la sunspot/sunspot#158.
- :+1: for eliminating kaminari dependency, that's a gotcha which has bitten me a few times now.
- :+1: for duck-typing detection and use of `find_in_batches`, really that's the 99% use case here.
- :+1: for the `GC.start` — I need that for sunspot asap. the counter argument here is that the GC will otherwise run when it needs to, and many gigs of used memory is not an issue if the process is due to merely exit soon.
- :+1: for some kind of progress bar. the `progress_bar` gem is nice to use when present, but probably not worth making a dependency.

…
- :-1: for passing multiple params to `puts` instead of interpolation (bike-shedding) :bike:
- :-1: for the scope stuff. not super clear on a quick skim. too hidden. how do I know whether and how to use this?

my 2¢ on a lazy saturday

jwaldrip:
@nz I am about to do another commit that should explain the idea a litte better. With sunspot I had to use the default AR scope, the idea here is that there should be a scope that the indexer should use that may not be used elsewhere in the app.

jwaldrip:
@karmi take a look per our conversation, let me know if this addresses the issues you were concerned about.

karmi:
Hi Jason, thanks for the update and all the care!

1/ I agree we should use `ansi/progress_bar` instead of the custom code. Another dependency though :/, pretty breakable one on top of that... I like how you make the `progress_bar` in the task optional, maybe we can make the dependency optional. Maybe I'm overly cautious :)

2/ Regarding the `rails environment`, I think it's best to leave it up to the user, really (Resque AFAIK does the same). But I agree that with the `if defined?(Rails)` check, this argument is weak...

3/ I am still opposed to importing multiple models in one go... It makes the code more complex, and I don't really see the utility here. Why not import the models separately? If we _really_ want to do this, then we should have a separate Rake task `tire:import:all`. I can see an argument for renaming the task to `tire:import:model` then, and prepare the ground for a generic `tire:import` task.

4/ If we go this route, I'm really starting to be concerned about testability of this code. When it's stashed inside the tasks.rb, there's no good way how to test it.

5/ The introspection in `model/import.rb` must be much more robust, not default to `paginate`, throw an error if the model doesn't have neither `find_in_batches` nor `paginate`, etc. It would be cool to research batch finders for most popular OxM libraries.

6/ I'm terribly sorry, but I don't like all the magic in `dependents` for models, loding based on `DIR` etc. As mentioned above, this is a separate Tire feature with `tire:import:model:all`.

7/ I like the idea behind `index_scope`, this issue turned out multiple times already, but I think we should extract it into a separate feature/issue. There's already lots of going on in this issue...

8/ Also note, that the whole section on importing in the README must be updated and rewritten.

lib/tire/model/search.rb
```diff
@@ -254,6 +259,7 @@ def initialize(instance)
       # A hook triggered by the `include Tire::Model::Search` statement in the model.
       #
       def self.included(base)
+        self.dependents << base
```
tpitale, line 262: DataMapper does something similar, but does not use mattr_accessor (Rails-only, maybe?). They call it descendants which matches the ancestors terminology quite well I think.

https://github.com/datamapper/dm-core/blob/master/lib/dm-core/model.rb#L46
https://github.com/datamapper/dm-core/blob/master/lib/dm-core/model.rb#L210

jwaldrip:
- Improving the rake task moved to #655 
- Adapter Specific importing moved to #656 
- DSL addition: index_scope moved to #657

lib/tire/model/search.rb
karmi, line 262: @tpitale I think checking [`MyClass.included_modules`](http://www.ruby-doc.org/core-2.0/Module.html#method-i-included_modules) is the only thing we need to do here, no need to track dependencies manually in the Tire code specifically -- unless I'm missing something obvious.

karmi:
Importing revamp added in #655 and #656, released as part of [v0.5.6](https://github.com/karmi/tire/compare/v0.5.5...v0.5.6).

PR #653, opened 2013-03-04
Remove the to_a; fixes data_mapper n+1 queries

tpitale:
When using Kaminari, whatever is returned from the scopes will behave as an array.

In the case of data_mapper this is especially important, because what gets returned also handles eager loading records. With the to_a in place, it's very likely that an n+1 query will be created during import.

### DIFF
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -76,7 +76,7 @@ namespace :tire do
       if defined?(Kaminari) && klass.respond_to?(:page)
         klass.instance_eval do
           def paginate(options = {})
-            page(options[:page]).per(options[:per_page]).to_a
+            page(options[:page]).per(options[:per_page])
           end
         end
       end unless klass.respond_to?(:paginate)


karmi:
Hi, I remember there have been some issues related to that on Mongoid, can you check it, please?

tpitale:
Oh wow, that would be sad. I wonder if that's a Kaminari problem, more than a Mongoid one. I'll investigate.

karmi:
Thanks! I'll verify it on a Mongoid app laters as well. Please ping me if I slip.

karmi:
Also, there's lots of going on in #652, where we wouldn't depend on pagination libraries for RDBMSes. Sadly, there's no common `find_in_batches` for ActiveModel.

tpitale:
Yeah, I saw that ticket too. I think, as long kaminari is still in that code, removing to_a would still be nice. Right now I use my own paginate, which is exactly the same as the one added by the rake task, except for the to_a :-)

tpitale:
Kaminari + Mongoid had lots of problems for a time, but there are currently no open issues that appear to mention anything related to this. I don't have an application that uses mongodb or mongoid handy, so I think that's all I can offer to the conversation for the moment. Let me know if you can't get to testing on a mongoid app of yours and I'll try to throw one together. Though, I doubt it would properly test mongoid pagination edge cases …

karmi:
@tpitale I'll verify it with a Mongoid + ActiveRecord apps laters!, I have some ready for this sort of manual testing.

karmi:
Hi, verified the injected `paginate` works for ActiveRecord and Mongoid importing. Merged & pushed, thanks!

PR #656, opened 2013-03-04
Adapter specific importing

jwaldrip:


jwaldrip:
Looking at this further, I wonder if this should just be added to index#import. Or have a more sophisticated way of managing this.

jwaldrip:

jwaldrip:
@karmi I think this needs more thought on my end. I will think deeper on it and provide a more elegant solution.

karmi:
@jwaldrip The basic idea is OK!, I'll take the code and take it from there, it needs only some reshaping.

karmi:
Hi Jason, I took the idea and added support for different importing strategies. Tested only manually, don't have the bandwith to add a full fledged test coverage at the moment.

PR #657, opened 2013-03-04
Added .index_scope

jwaldrip:


jwaldrip:

karmi:
Been thinking about this for a while, and I still like the idea behind it, but not the exact implementation. The scope would be picked up only for batch importing, not when storing records individually, which someone might find surprising. A feature like this would have to documented in detail, on top of that.

From a rather philosophical standpoint, I don't see the library as trying magically support any use case and offer API/DSL for any imaginable scenario. I'd like it to provide very clean and convenient building blocks, so anybody can very easily write eg. their own import routines, in Ruby, using all the power of the language.

In other words, I see more benefit in having a nice chapter in documentation elaborating on how to achieve this ("automatically restrict search results to a certain ActiveRecord scope"), instead of providing an API how to do that directly. 

So, let's imagine we have this system, where we want to index only published articles. All the code needed to customize Tire for this would fit on a handful of lines:

``` ruby
# Model

  include Tire::Model::Search
  ...
  scope :published, where(:published => true)

  after_commit  { tire.update_index if published? }
  after_destroy { tire.update_index if published? }
  ...
end

# Rake task
desc "Import articles"
task :import do
  index = Tire::Index.new( ENV['INDEX'] || Article.tire.index.name )

  if ENV['FORCE']
    index.delete and index.create( :mappings => Article.tire.mapping_to_hash, :settings => Article.tire.settings )
  end

  Article.published.find_in_batches do |batch|
    puts "Importing #{batch}"
    index.import batch
  end
end
```

What do you think?

---

Regarding the example with `includes(:comments)`, that's already supported (bcbe4dd):

```
rake environment tire:import PARAMS='{:include => ["comments"]}' CLASS=Article
```

Though I'm not sure if the ActiveRecord API still supports it.

jwaldrip:
@karmi Well, I have been digging even deeper into this. It is great that the import rake task supports this. But we have a very complex case. Here is what the code looks like. Also , I added some automagical includes for the active record api. As you may be able to determine from below is as the rake task supports some simple includes. A much more complex solution is very tedious.

``` ruby
module Tire::IndexingAdditions
  extend ActiveSupport::Concern

  def to_indexed_json
    to_json mapping_to_json_options
  end

  def mapping_to_json_options(mapping = self.class.tire.mapping_to_hash)
    options = mapping.reduce({}) do |options, (k, v)|
      if v.has_key?(:properties)
        options[:only]    = []
        options[:methods] = []
        options[:include] = {}
        v[:properties].each do |k, v|
          if v.has_key?(:properties) && self.class.reflect_on_association(k)
            options[:include][k] = mapping_to_json_options({ k => v })
          elsif attribute_names.include? k.to_s
            options[:only] << k
          else
            options[:methods] << k
          end
        end
      end
      options
    end
    options.update({ root: false })
  end

  module ClassMethods

    def mapping_to_arel(arel=self.includes, mapping=self.tire.mapping_to_hash)
      mapping.reduce(arel) do |arel, (k, v)|
        arel = if v.has_key? :properties
                 v[:properties].reduce(arel) do |arel, (k, v)|
                   if reflect_on_association k
                     arel = case arel
                            when ActiveRecord::Relation
                              arel.includes(k => mapping_to_arel([], { k => v }))
                            when Array
                              inner_mapping = mapping_to_arel([], { k => v })
                              append = inner_mapping.present? ? { k => inner_mapping } : k
                              arel << append
                            end
                   end
                   arel
                 end
               end
        arel
      end
    end
  end
end

module MedicalFacilityIndexer
  extend ActiveSupport::Concern
  include Tire::IndexingAdditions

  included do
    include Tire::Model::Search

    tire do

      index_scope do
        mapping_to_arel.
          includes(appointment_books: { appointment_reasons: {}, appointment_setting: {} }).
          includes(:premium_profile).
          includes(:administrative_area).
          includes(physician_appointment_books: { appointment_reasons: {}, appointment_setting: {} }).
          includes(:medical_facility_pre_registration).
          includes(physician: [:account, :ratings, :premium_profile, :languages, { medical_facility_group: :premium_profile }]).
          includes(medical_facility_group: { premium_profile: {} })
      end

      mapping do

        # Basic Properties
        indexes :name, type: 'string'
        indexes :updated_at, type: 'date'

        # Address
        indexes :address
        indexes :physical_state
        indexes :physical_city
        indexes :physical_zip
        indexes :location, type: "geo_point", as: "[lat, lng]", lat_lon: true, geohash: true

        # Properties
        indexes :is_premium?, type: 'boolean'
        indexes :includes_photo?, type: 'boolean'
        indexes :includes_website_link?, type: 'boolean'

        # Home Health Care
        indexes :service_area, type: "nested" do
          indexes :options, type: "string"
        end

        # Widget Ids
        indexes :widgets, type: 'nested' do
          indexes :id, type: "integer"
        end

        # Appointment Setting
        indexes :pre_registration_enabled?, type: 'boolean'
        indexes :is_bookable?, type: 'boolean'
        indexes :appointment_book_enabled?, type: 'boolean'
        indexes :appointment_reasons, type: 'string'

        # Physician
        indexes :has_physician?, type: 'boolean'
        indexes :physician, type: "nested" do
          indexes :id, type: "integer"
          indexes :gender, type: "string"
          indexes :experience, type: "string"
          indexes :average_rating, type: "integer"
          indexes :is_premium?, type: "boolean"
          indexes :medical_specialties, type: "nested" do
            indexes :id, type: "integer"
            indexes :name, type: "string"
          end
          indexes :languages, type: "nested" do
            indexes :id, type: "integer"
            indexes :name, type: "string"
          end
        end

        # Categories
        indexes :medical_facility_categories, type: "nested" do
          indexes :id, type: "integer"
          indexes :category, type: "string"
          indexes :slug, type: "string"
        end

        # Sub Categories
        indexes :medical_facility_sub_categories, type: "nested" do
          indexes :id, type: "integer"
          indexes :name, type: "string"
        end

        # Account
        indexes :account, type: "nested" do
          indexes :id, type: "integer"
          indexes :name, type: "string"
        end

        # Group
        indexes :medical_facility_group, type: "nested" do
          indexes :id, type: "integer"
          indexes :name, type: "string"
        end

      end
    end
  end

end
```

karmi:
Jason, first, nothing makes me more happy then to see Tire being used in a more complex setup like this!

Second, my vision for Tire or its successors is precisely this: provide building blocks to assemble a complicated domain like that. At the moment, Tire has a lot of quirks and lacks support to make this really a joy.

I understand what you're saying, but in my eyes, it _supports_ the argument I'm trying to make -- that eg. the bundled import task is merely a starting point, a convenience, and that it shouldn't be complicated or convoluted to write your own implementation, for your own domain. Notice how in this area, I always say “just don't include the `Tire::Model::Callbacks` module then”. The _mechanics_ of indexing are quite simple, as outlined in the snippet I've posted. Of course, when you have a complicated _domain_ to work with, it's not easy -- but that's an intrinsic aspect.

As a side note, the code example you've posted illustrates something more problematic in Tire, and that is the whole confusion between `mapping` and `to_indexed_json`, which is something I'd like to see solved much more then any ActiveRecord scoping issues and such. It's evident how hard are you coding _against_ the library, and that is something which I don't like to see; the building blocks are not matching well here.

PR #658, opened 2013-03-04
Fix memory leaks on large imports

jwaldrip:
GC.start after consecutive batch imports causes a memory leak. While importing 6 million records, the stack grew from 200mb to ~5GB. This fixes that issue.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -246,11 +246,12 @@ module Tire
         when method = options.delete(:method)
           options = {:page => 1, :per_page => 1000}.merge options
           while documents = klass_or_collection.send(method.to_sym, options.merge(:page => options[:page])) \
-                            and documents.to_a.length > 0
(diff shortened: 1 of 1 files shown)


lib/tire/index.rb
```diff
@@ -246,11 +246,12 @@ def import(klass_or_collection, options={})
         when method = options.delete(:method)
           options = {:page => 1, :per_page => 1000}.merge options
           while documents = klass_or_collection.send(method.to_sym, options.merge(:page => options[:page])) \
-                            and documents.to_a.length > 0
+                            && documents.to_a.length > 0
```
jwaldrip, line 249: Also, && is better to use than "and" in conditional arguments.

jwaldrip:

karmi:
I'm not a huge fan of manually triggering GC, but pulling this in, we'll see how it turns out. Thanks!

PR #663, opened 2013-03-05
Merge fields in with source if they are present

jwaldrip:
When using script_field with a source present, it is very hard to access the values in the item object. This solves this issue with the caveat that specified fields will override their source values.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -106,12 +106,13 @@ module Tire
           hits.map do |h|
             document = {}
 
-            # Update the document with content and ID
-            document = h['_source'] ? document.update( h['_source'] || {} ) : document.update( __parse_fields__(h['fields']) )
-            document.update( {'id' => h['_id']} )
+            # Update the document with fields and or source
+            document.update h['_source'] if h['_source']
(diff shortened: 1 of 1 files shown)


jwaldrip:

karmi:
I'm sorry but I can't see how “it is very hard to access the values in the item object”. Would it be possible to provide a code example, or preferably a test case?

jwaldrip:
When you load the instance with the query this works fine. Where it doesnt work is where you rely on the source from ES only. For example:

``` ruby
results = MedicalFacility.tire.search page: (params[:page] || 1), per_page: (params[:per_page] || 10) do |search|
  search.query { all }
  search.fields "_source"
  search.script_field :distance, script: "doc['location'].distance(#{lat_lng.join(',')})"
  # ...
end

results.first.distance
-> nil

```

As you can see without the fix, the script field distance returns nil, with the fix, I get the proper value.

karmi:
Ah, sure. Can you add a unit test for this change?

mattiassvedhem:
ping, @jwaldrip I've added the unit test to your fork, can you take a look? :)

PR #673, opened 2013-03-13
change #find to #find_by_ids

iwarshak:
Use find_by_ids when loading the collection. If a record was deleted, but ES has not been updated yet, #find would throw a ActiveRecord::RecordNotFound exception.
# find_by_ids returns only the records that are available

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -148,7 +148,7 @@ module Tire
       end
 
       def __find_records_by_ids(klass, ids)
-        @options[:load] === true ? klass.find(ids) : klass.find(ids, @options[:load])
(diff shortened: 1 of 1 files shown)


karmi:
Hi, the trouble here is that there's no common `find_by_ids` fro ActiveModel, and Tire tries to work with many other OxM frameworks apart from ActiveRecord. I think there's been some discussion about it in the issues already, the main reason for the refactoring in bac091a and introducing of `__find_records_by_ids` was to provide a hook point for people needing to do this.

I think the real solution would be to abstract this behaviour even further, and provide and extension (possibly autoloaded) for ActiveRecord, which would use `find_by_ids`...

karmi:
Closing in favor of a more robust solution, which provides an extensions when the host class inherits `ActiveRecord::Base` and automatically changes `__find_records_by_ids` in the outlined sense -- is that OK?

PR #684, opened 2013-03-21
using find_all_by_id to prevent record not found errors when there is a missing record. 

rahulag:
Using find_all_by_id to prevent record not found errors when there is a missing record.
Always returns an array even if there is a single object to find.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -148,7 +148,7 @@ module Tire
       end
 
       def __find_records_by_ids(klass, ids)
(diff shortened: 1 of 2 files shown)


karmi:
@rahulag Please search the issues for `find_by_ids`, eg. #673, to see the reasoning why this is not a plausible solution.

rahulag:
I am using find_all_by_ids.. this doesn;t throw error when record is not there and returns an array even when there is a single record to be fetched. 

Also using Array.new in terms query. This is for backward compatibility if a non array value is used for terms query.

PR #685, opened 2013-03-23
Fix Other Clients link in README

leschenko:
Other Clients link returns 404
Changed to [http://www.elasticsearch.org/guide/clients/](http://www.elasticsearch.org/guide/clients/)


karmi:
Thanks!, updated as part of larger changes in efb9c14.

PR #699, opened 2013-04-09
Allow for specifying version in Tire::Index#store

cjbottaro:
Allow for the version param as specified here http://www.elasticsearch.org/guide/reference/api/index_/


karmi:
Added a test and merged, thanks!

PR #700, opened 2013-04-11
added read_attribute_for_serialization to result item

lasssim:
I added the method **read_attribute_for_serialization** to the result item to support active model serializers.

### DIFF
--- a/lib/tire/results/item.rb
+++ b/lib/tire/results/item.rb
@@ -34,6 +34,10 @@ module Tire
         @attributes[key.to_sym]
       end
 
+      def read_attribute_for_serialization(key)
+        @attributes[key.to_sym]
+      end
+
       def id
         @attributes[:_id]   || @attributes[:id]
       end
--- a/lib/tire/results/item.rb
+++ b/lib/tire/results/item.rb
@@ -33,10 +33,9 @@ module Tire
       def [](key)
         @attributes[key.to_sym]
       end
+      
+      alias :read_attribute_for_serialization :[]
 
-      def read_attribute_for_serialization(key)
-        @attributes[key.to_sym]
-      end
 
       def id
         @attributes[:_id]   || @attributes[:id]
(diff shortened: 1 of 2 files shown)


karmi:
Can you please check with https://github.com/karmi/tire-contrib/pull/15 and https://github.com/ResumUP/tire_am_serializers, and also add test cover for the feature?

lasssim:
hi,

sorry for the missing test. i just added them.

the tire_am_serializers gem only helps with autodetecting a serializer as far as i can tell from skimming the code.

active model serializers uses the read_attribute_for_serialization method to retrieve attributes from the objects it serializes. this is the part from active model serializers (activemodel-3.2.11/lib/active_model/serialization.rb:100)

``` ruby
      # Hook method defining how an attribute value should be retrieved for
      # serialization. By default this is assumed to be an instance named after
      # the attribute. Override this method in subclasses should you need to
      # retrieve the value for a given attribute differently:
      #
      #   class MyClass
      #     include ActiveModel::Validations
      #
      #     def initialize(data = {})
      #       @data = data
      #     end
      #
      #     def read_attribute_for_serialization(key)
      #       @data[key]
      #     end
      #   end
      #
      alias :read_attribute_for_serialization :send

```

br, simon

mulderp:
Hm.. I can't confirm this commit to make my ActiveModelSerializer work. I am using:

```
    movies = Movie.search(params[:q], :load => true)
    render :json => movies, :each_serializer => MovieSerializer
```

But get a:

```
   NoMethodError (undefined method `<=' for false:FalseClass)
```

Let me know, if I can test some variation.

karmi:
@lister Sorry for the ridiculous delay! Merged the patch and added some tests for ActiveModel::Serializers (d2263fb)

@mulderp Could you try again? Notice (d2263fb), that you need to specifically include the support in `Tire::Results::Item`, if you'd need to serialize those. Your models should work fine I guess -- please paste the fuller backtrace for the error.

Issue #283, opened 2012-03-20
Add proper DSL for geo filtering


karmi:
While the geo filtering is [currently possible to do with Tire](https://github.com/karmi/tire/issues/140#issuecomment-2568675), it would be nice to have a more convenient DSL support for geo features:

``` ruby
Tire.search('venues') do
  query  { all }
  geo    { distance :location, '200km', [40, -71] }
end

Tire.search('venues') do
  query  { all }
  geo    { polygon :location, [40, -70], [30, -80], [20, -90] }
end

Tire.search('venues') do
  query  { all }
  geo    { bounding_box :location, :top_left => [40, -70], :bottom_right => [30, -80] }
end
```

PR #701, opened 2013-04-11
Add support for geo distance facets

robinbrandt:
This pull request adds implementation and tests for geo distance facets to tire, see http://www.elasticsearch.org/guide/reference/api/search/facets/geo-distance-facet for more details.

Thanks for your work on tire. We use it at Stuffle (http://stuffle.it) and it's great how little effort was needed to integrate ElasticSearch.

Cheers,
Robin

### DIFF
--- a/lib/tire/search/facet.rb
+++ b/lib/tire/search/facet.rb
@@ -2,7 +2,7 @@ module Tire
   module Search
 
     #--
-    # TODO: Implement all elastic search facets (geo, histogram, range, etc)
+    # TODO: Implement remaining elastic search facets (date histogram)
     # http://elasticsearch.org/guide/reference/api/search/facets/
     #++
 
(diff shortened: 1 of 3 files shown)


robinbrandt:
@karmi any thoughts?

karmi:
@robinbrandt I like it, thanks!, will try to process it as soon as possible...

karmi:
@robinbrandt By the way, what do you think about #283? Do you use geo filtering as well?

robinbrandt:
We don't use it yet but probably in a couple of hours ;) I have to play around with it a bit more but my first expression was that geo distance is just another filter and the filtering DSL is flexible enough to support it so there's no need for another special geo syntax. 

But I'll give it a try. When I find out I was wrong, I'll grab the ticket, ok?

robinbrandt:
Hi @karmi !

Any idea if this could make it into the next tire release?

karmi:
@robinbrandt Sorry it took so long! Merged and pushed, thanks for a great patch! I'll include it in the next release, once I solve other hanging issues, will release probably tommorrow.

robinbrandt:
No problem, thanks for your work.

PR #702, opened 2013-04-11
Change #find to #where to prevent ActiveRecord::RecordNotFound

jeyb:
As per @karmi's comments on https://github.com/karmi/tire/pull/673 I've only used where when the class inherits ActiveRecord::Base.

Thoughts @karmi?

### DIFF
--- a/test/unit/results_collection_test.rb
+++ b/test/unit/results_collection_test.rb
@@ -235,7 +235,7 @@ module Tire
           ActiveRecord::Schema.define(:version => 1) { create_table(:active_record_articles) { |t| t.string(:title) } }
           model = ActiveRecordArticle.new(:title => 'Test'); model.id = 1
 
-          ActiveRecordArticle.expects(:find).with([1]).returns([ model] )
+          ActiveRecordArticle.expects(:where).with(id: [1]).returns([ model])
 
           Results::Collection.new(response, :load => true).each_with_hit do |result, hit|
             assert_instance_of ActiveRecordArticle, result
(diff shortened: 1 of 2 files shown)


test/unit/results_collection_test.rb
```diff
@@ -235,7 +235,7 @@ class ResultsCollectionTest < Test::Unit::TestCase
           ActiveRecord::Schema.define(:version => 1) { create_table(:active_record_articles) { |t| t.string(:title) } }
           model = ActiveRecordArticle.new(:title => 'Test'); model.id = 1
 
-          ActiveRecordArticle.expects(:find).with([1]).returns([ model] )
+          ActiveRecordArticle.expects(:where).with(id: [1]).returns([ model])
```
robinbrandt, line 238: You're using Ruby 1.9 Hash syntax here. I don't know the general policy here (I just added my first pull request myself) but the other tests (at least the unit tests) use the 1.8 Hash syntax. But from some older commits and the documentation you can see that 1.8 compatibility is desired.

karmi, line 238: Yes. Ruby 1.8 compatibility of the whole gem is in fact _required_, the only place where we commonly use Ruby 1.9 syntax is integration tests (easier on the eyes, most people are on Ruby 1.9, Hash preserves ordering, etc). The unit tests have to use the Ruby 1.8 syntax, because we run the unit tests suite on Ruby 1.8 on Travis (and of course locally).

karmi:
@jeyb See also discussion in #602, please.

jeyb:
@karmi @robinbrandt I can definitely update to Ruby 1.8 syntax.

@karmi If you can point out where I can find a list of supported ORMs I can work on a `Strategy` implementation like done here https://github.com/karmi/tire/commit/efb9c14. Is this already being worked on? Don't want to dup effort. But this fix is critical for my project so I do wan't to work on it soon.

brupm:
@karmi bump

karmi:
Sorry it hangs in there like this for so long. Any news on this in the meantime? I think you should synchronize with @timoschilling to work out some elegant solution which would pick correct strategy for `__find_records_by_ids` for ActiveRecord, Mongoid, etc. Let's keep the current implementation as default.

karmi:
Just wondering about the state of this ticket -- I'm all for exposing ORM-specific finder methods in the `__find_records_by_ids`, but would like to have it done properly, so we can extend it, covered it by tests, etc. Currently, for a homogenous application, people can just override the method, but it would be cool to have a proper out-of-the-box support here...

jeyb:
Didn't get a chance to work on this till a few minutes ago. How does this look? https://github.com/karmi/tire/pull/810

karmi:
Closing in favour of #810.

PR #709, opened 2013-04-18
Add 'strategy' option to bulk import

leschenko:
During bulk import for active_record models sometimes it is important to eager load associations, which can't be done with default strategy for active_record models.

### DIFF
--- a/lib/tire/model/import.rb
+++ b/lib/tire/model/import.rb
@@ -26,6 +26,9 @@ module Tire
       #
       module Strategy
         def self.from_class(klass, options={})
+          if options[:strategy]
+            return const_get(options[:strategy]).new(klass, options)
+          end
           case
           when defined?(::ActiveRecord) && klass.ancestors.include?(::ActiveRecord::Base)
             ActiveRecord.new klass, options
--- a/lib/tire/model/import.rb
+++ b/lib/tire/model/import.rb
@@ -26,9 +26,7 @@ module Tire
       #
       module Strategy
         def self.from_class(klass, options={})
-          if options[:strategy]
-            return const_get(options[:strategy]).new(klass, options)
-          end
+          return const_get(options[:strategy]).new(klass, options) if options[:strategy]
           case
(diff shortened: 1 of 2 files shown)


karmi:
I like the injection of the strategy!, just a question if it wouldn't be possible to pass some options/arel blocks to the strategy down the wire? I was aware that we lost the ability to eager load associations when merging this, and didn't see a good way how to implement that.

Couple of notes:

1/ could you add a unit test for this?

2/ maybe compress the code into one line:

```
return const_get(options[:strategy]).new(klass, options) if options[:strategy]
```

3/ Wouldn't the `import` method be a better location for this? It's kinda confusing to return the injected strategy from the `Strategy.from_class` method...?

leschenko:
There is a lot of models with custom `paginate` method for old tire bulk import scheme. So it would be nice to have the ability to use them after tire update via WillPaginate strategy.

1/ I didn't find a good place for test because import strategies didn't covered with unit tests yet
2/ definitely!
3/ may be

May be this approach isn't elegant, but I didn't find better way yet.

karmi:
It's a very good solution. People will just need pointers how to implement their strategy and use it. An integration test would be cool even for this -- I like to point people to working code instead to docs... What do you think?

karmi:
@leschenko Hi, any news on this? I think an integration test as docs + plus the suggested changes would make this quite helpful for people who need to customize the import behaviour.

leschenko:
Hello, sorry for delay.
I added unit test for this option and compressed the code into one line.
I have question about integration test: is it a test for rake tasks or Tire::Tasks::Import module?

karmi:
@leschenko Merged and pushed, thanks!

Example:

``` ruby
class MyCustomImportStrategy
  include Base
  def import &block
    klass.
      where(published: true).
      joins(:comments).
      find_in_batches(:batch_size => options[:per_page]) do |batch|
        index.import batch, options, &block
      end
    self
  end
end
```

PR #710, opened 2013-04-21
Add JavaScript Runtime to example

dariusgm:
This will prevent the example to thow errors.

### DIFF
--- a/examples/rails-application-template.rb
+++ b/examples/rails-application-template.rb
@@ -107,7 +107,7 @@ puts        '-'*80, ''; sleep 1
 
 gem 'tire', :git => 'git://github.com/karmi/tire.git'
 gem 'will_paginate', '~> 3.0'
-gen 'therubyracer'
+gem 'therubyracer'
 
 git :add => '.'
 git :commit => "-m 'Added gems'"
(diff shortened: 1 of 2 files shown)


karmi:
Hey, this is strange, shouldn't it be handled by Rails itself?

dariusgm:
I think the example App should run instantly. As I downloaded it, it didn't work. After adding the run time it worked. I don't know if this should be handled by the user or be inside the example.

karmi:
Yes, the example app should definitely run out of the box. But it does on my machine:

```
$ gem list therubyracer

*** LOCAL GEMS ***


$ rails new searchapp -m https://raw.github.com/karmi/tire/master/examples/rails-application-template.rb
...
=> Rails 3.2.13 application starting in development on http://0.0.0.0:3000
```

What error do you receive?

dariusgm:
Using Ruby 2.0-p0 as Run time Environment create this error while launching the rails server. The Rake Task Abort with the message:
<code>Could not find a JavaScript runtime. See https://github.com/sstephenson/execjs for a list of available runtimes. (...)</code> - same while indexing the database.

karmi:
Actually, on Ruby 2.0 I'm getting an error related to https:

```
ruby/2.0.0/net/http.rb:917:in `connect': SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (OpenSSL::SSL::SSLError)
```

I wouldn't like to add JS runtime to the example, since this is something Rails should take care of. Closing this for the moment, and will have a look on how to solve the SSL issues on 2.0...

PR #713, opened 2013-04-24
Pooled curb v1

brupm:
- Uses https://github.com/jugend/common-pool to allow for high-performance HTTP requests for a multi-threaded environment.
- Properly handles `Curl::Err::MultiBadEasyHandle`
- I will write tests if you think there's validity in this PR.

### DIFF
--- a/lib/tire/http/clients/pooled_curb.rb
+++ b/lib/tire/http/clients/pooled_curb.rb
@@ -1,0 +1,114 @@
+require 'common_pool'
+require 'curb'
+
+module Tire
+
+  module HTTP
+
+    module Client
+
+      class PooledCurbDataSource < CommonPool::PoolDataSource
+
+        def create_object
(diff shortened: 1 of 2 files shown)


karmi:
Hi, great! Can you please squash the commits into one and also remove the dependency from the gemspec?

brupm:
@karmi all done. The dependency appears to be required, otherwise all will not boot. @karmi Am I missing something?

karmi:
> The dependency appears to be required, otherwise all will not boot

It's the same thing as with eg. the Faraday client -- you need to require the dependency externally, eg. in an application initializer, etc. On a second thought, can you please move the pull request from here to [tire-contrib](https://github.com/karmi/tire-contrib)?

brupm:
https://github.com/karmi/tire-contrib/pull/18

PR #716, opened 2013-04-26
Allow wrapper option to be passed to Tire.search for all scenarios.

mattiassvedhem:


mattiassvedhem:
It's not a very elegant solution, improvements are welcome.

karmi:
@mattiassvedhem I'm sorry to keep it hanging like this for a long time, I'll have a look. The problem is that the "passing Hashes to search" usecases are really fragile in the current codebase... (Bad initial design decisions :)

mattiassvedhem:
@karmi No worries! Yeah, I can see that, is there another way to augment a search instance with unsupported ElasticSearch features? Or do you feel that it should it be made to work another way? Maybe there could be a method allows for direct augmentation of custom hashes, that then gets merged in `to_hash`, but that might be tricky.

karmi:
Briefly verified the code and it seems be working fine, pushed. Thanks!

PR #718, opened 2013-05-01
Make Tire::Results::Collection.to_ary return an array

brandones:
This commit fixes the fact that calling to_ary() on a Tire::Results::Collection object returns another Tire::Results::Collection object... which is not an array.

### DIFF
--- a/lib/tire/results/collection.rb
+++ b/lib/tire/results/collection.rb
@@ -58,7 +58,7 @@ module Tire
       alias :[] :slice
 
       def to_ary
-        self
+        results
       end
 
       def as_json(options=nil)
--- a/test/unit/results_collection_test.rb
+++ b/test/unit/results_collection_test.rb
@@ -79,6 +79,13 @@ module Tire
         assert_equal 1.0, collection.max_score
       end
 
+      should "return a Ruby array" do
+        collection = Results::Collection.new(@default_response)
+        to_ary_result = collection.to_ary
+        array = Array.new
+        assert to_ary_result.is_a? array
+      end
+
       context "serialization" do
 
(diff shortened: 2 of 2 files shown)


karmi:
Hi, is there any specific problem with the implementation? If I recall correctly, `to_ary` is used in Rails' views, have you verified it works correctly?

brandones:
I see the fact that "to_ary" does not return an array as a specific problem with the implementation (or maybe the design).

The change doesn't appear to have broken my app nor tire's tests. But many tests errored out on my machine, so I'm not totally sure.

karmi:
@brandones The `to_ary` modifier is used extensively in Rails views. There's no test for them in Tire test suite.

karmi:
@brandones Any new thoughts about this? Have you tested the change within a Rails application? (You can generate one with the included template)

brandones:
The change seems not to have affected either the website I've been building or the Tire template site ("tired").

As an additional note, I found while working with the template site that inserting

```
artarr = @articles.to_ary
puts artarr
```

at `app/controllers/articles_controller.rb:17` would throw 

```
TypeError in ArticlesController#search
can't convert Tire::Results::Collection to Array (Tire::Results::Collection#to_ary gives Tire::Results::Collection)
```

which, with the change, it no longer does.

ivobenedito:
Started using @brandones PR in a project with tire + rabl and everything seems to work fine.
Previously, wasn't able to render tire collections with rabl due to this bug.

karmi:
Hi, checked it briefly against example apps, seems to be working fine -- simplified the test and pushed, thanks!

PR #719, opened 2013-05-01
Compacted line of get the type from the Hash instance 

miry:
simple changes in `get_type_from_document`


karmi:
Is this really a simplification?

karmi:
@miry Unless you object strongly, I'd keep the current code, since it's arguably more readable / simple to understand?

miry:
ok

karmi:
Thanks!

PR #720, opened 2013-05-01
Removed indents in the code blocks of the README file.

miry:


karmi:
The problem is that other Markdown parsers then don't understand these sections as code blocks.

miry:
@karmi can you provide one of this parsers. 

I used these to test: 
- http://daringfireball.net/projects/markdown/dingus
- http://markable.in/editor/
- Rubymine plugin

karmi:
For instance the Textmate bundle doesn't handle it right, also:

```
File.open('index.html', 'w') do |f| f << RDiscount.new(File.read('README.markdown')).to_html end
```

doesn't render the "Github markdown" blocks right without initial indentation. What's the reasoning for this change?

miry:
hmm, it works like all other tools https://gist.github.com/miry/5497817. Used `rdiscount-2.0.7.2.gem`.

The reason is when I copy some of this examples to `irb` there are some extra spaces.
And it is "offended the eye" ("cut the eye").

not critical

Issue #339, opened 2012-05-08
Nested to_hash in Tire::Results::Item


thedaniel:
When Tire::Results::Item wraps a nested hash like

```
{
 username: bob,
 avatars: {
   big: 'http://foo',
   small: 'http://bar
  }
}
```

both the outer hash and inner hash (`avatars` in this example) are cast to Tire::Results::Item. This is awesome for dot notation access to nested hashes.

Unfortunatey, if you want to manually serialize this with e.g. `to_json`, only the outer hash is recast to Hash, leaving the inner has still as an Item. This causes to_json to crash like:

```
ruby-1.9.2-p180 :021 > h.to_json
ArgumentError: wrong number of arguments (2 for 1)
    from /Users/daniel/.rvm/gems/ruby-1.9.2-p180@ng2/gems/tire-0.4.2/lib/tire/results/item.rb:75:in `to_json'
```

Basically, i think that to_hash and to_json should recursively un-cast from Item back to Hash, not just do the outermost layer.

thedaniel:
After looking at my data that caused the problem, i think it might not be nested hashes, but Items inside arrays, like

```
{
  foo: 'bar',
  baz: [Item(), Item()],
}
```

karmi:
Hmm, this sounds like a nasty bug. Could you have a look into unit tests for `Tire::Results::Item` if you'd be able to write a failing test?

thedaniel:
whoops, just saw this - i had notifications turned off for some reason. i will make some time to try to write a failing test this week sometime

azitabh:
Any update on this?

karmi:
Should be fixed in karmi/tire@4f2cbd4

karmi:
@azitabh @thedaniel Thanks for your feedback, hope your problem is solved. Closing for now.

PR #722, opened 2013-05-02
Fix incomplete nested to_hash functionality

konradreiche:
Issue #339 addressed the lacking functionality for the `to_hash` to work in a recursive fashion. This was fixed, however, I noticed it does not work if `Tire::Results::Item` contains an array with `Tire::Results::Item` objects. In this case the array is just skipped.

```
<Item title:   "Test", author: <Item name: "Kafka">,
      awards:  <Item best_fiction: <Item year: "1925">>,
      reviews: [<Item stars: 5, comment: "great">,
                <Item stars: 3, comment: "decent">]
```

becomes

```
{:title =>   "Test", :author => {:name => "Kafka"},
 :awards =>  {:best_fiction => {:year => "1925"}},
 :reviews => [<Item stars: 5, comment: "great">,
              <Item stars: 3, comment: "decent">]}
```

and not as expected

```
{:title   => "Test", :author => {:name => "Kafka"},
 :awards  => {:best_fiction => {:year => "1925"}},
 :reviews => [{:stars => 5, :comment => "great"},
              {:stars => 3, :comment => "decent"}]}
```

I have extended the unit test case to reveal this problem and added a fix. I hope this helps, at least I need this, as I was expecting this behavior. I am also curious if there is an alternative, more elegant, implementation to my solution.

### DIFF
--- a/lib/tire/results/item.rb
+++ b/lib/tire/results/item.rb
@@ -60,7 +60,11 @@ module Tire
 
       def to_hash
         @attributes.reduce({}) do |sum, item|
-          sum[ item.first ] = item.last.respond_to?(:to_hash) ? item.last.to_hash : item.last
+          if item.last.is_a?(Array)
+            sum[ item.first ] = item.last.map { |item| item.respond_to?(:to_hash) ? item.to_hash : item }
+          else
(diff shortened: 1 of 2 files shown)


karmi:
Konrad, this sounds highly plausible! I don't think there's a more elegant/concise way how to check if the item being worked on is an Array, this solution is readable.

PR #726, opened 2013-05-10
Should be `tire`, not `index` (README)

justinko:


brupm:
It should be `index`

karmi:
Actually, both are possible :) But in the context, the change suggested by @justinko is correct, since we want to use the newly added ["polymorphic" import](https://github.com/justinko/tire/blob/master/lib/tire/model/import.rb) strategies.

brupm:
I stand corrected, didn't know about polymorphic import

justinko:
Well let's merge this muthafucka! (Samuel L. Jackson voice)

karmi:
Actually, when reading the passage in context, `Article.index.import` is right -- it communicates clearly that every `Tire::Index` object has an `import` method, and you can pass a collection (`Enumerable`) to it.

In contrast, `Article.import`, will try to pick up a strategy for handling loading the records.

Bottom line, current wording is correct, but we should advertise the strategies more in the README, not just say “depending on the setup of your model ...”.

PR #524, opened 2012-11-17
Cluster Health and Optimize additions

marckysharky:

PR #731, opened 2013-05-17
Cluster commands

pzac:
99% of this code comes from https://github.com/marckysharky/tire. He made pull request [#524](https://github.com/karmi/tire/pull/524) some months ago; this one merges in the latest commits and adds a Procfile. Please let me know if this feature should be implemented in `tire-contrib` rather than here.


karmi:
Hi, I'm sorry, but I wouldn't like to add new APIs to Tire at this stage. It could theoretically be added as a part of tire-contrib, but I'd wait a bit with it.

PR #734, opened 2013-05-20
support custom filters score

fcheung:
Adds support for the custom filters score query type ( http://www.elasticsearch.org/guide/reference/query-dsl/custom-filters-score-query/ )

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -109,6 +109,13 @@ module Tire
         @value
       end
 
+      def custom_filters_score(options={}, &block)
+        @custom_filters ||= CustomFiltersScoreQuery.new(options)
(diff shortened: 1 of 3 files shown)


brupm:
This is already in tire-contrib https://github.com/karmi/tire-contrib/blob/master/lib/tire/queries/custom_filters_score.rb @fcheung

karmi:
@fcheung I've added the patch, and also added a more DSL-ish variant. I think the tire-contrib version is superior, as @brupm noted :), can you check if it satisfies your use case? I'll move it over from contrib to core I guess.

fcheung:
I hadn't spotted the tire-contrib version - It hadn't ocurred to me to look there. Looks fine to me.

karmi:
@fcheung Embarassingly enough, I forgot as well :) Will create a mashup of these and push it.

PR #739, opened 2013-05-24
Small error in documentation

jurgens:


karmi:
Hey, actually the doc is correct, observe:

``` ruby
s = Tire.search('articles') do
  facet 'tags' do
    terms :tags
  end
end

p s.facets
#=> {"tags"=>{:terms=>{:field=>:tags, :size=>10, :all_terms=>false}}}

p s.results.facets
#=> {"tags"=>{"_type"=>"terms", "missing"=>0, "total"=>3, "other"=>0, "terms"=>[{"term"=>"ruby", "count"=>1}, {"term"=>"python", "count"=>1}, {"term"=>"java", "count"=>1}]}}
```

In the first case, you get the facet "definition" for the request, in the second case, you get the actual response...

PR #743, opened 2013-05-30
Added support for geo_shape query

chrishale:
Use like so...

``` ruby
Model.search do
    query do
        boolean do
            must { geo_shape :place_location, { :shape => { :type => 'point', :coordinates => [23,42] }, :relation => 'intersects' } }
        end
     end
end
```

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -109,6 +109,11 @@ module Tire
         @value
       end
 
+      def geo_shape(field, value)
+        @value = { :geo_shape => { field => value }}
+        @value
+      end
+
       def to_hash
         @value
       end


karmi:
Hi, unfortunately, I can't pull it in just like that :) For a patch like this, a proper unit test and a working, non-synthetic integration test are required... Also, I was thinking about a more DLS-ish approach to geo stuff, as outlined in #283, what do you think about that?

chrishale:
Cool, sounds good. It would be nice if you could nest it in a query or a filter. I don't personally think it needs abstracting, but if the documentation is clear enough - some kind of DSL would be good. I'm using the geo filters, queries and facets pretty extensively in a current project, I might have a crack at another pull request in the next few weeks... or months ;)

karmi:
@chrishale If it works for you as it is, no rush. What would be cool though, if you could add it to the https://github.com/karmi/tire/wiki/How-to-work-with-locations-(geo)-in-Tire wiki page.

Even more cool would be a wiki page or a README section about how easy it is to inject/monkeypatch/extend the Tire query class with a one-off method like this, when certain functionality such as `geo_shape` is missing.

PR #754, opened 2013-06-10
Add basic support for the regexp query

antekpiechnik:
As described here:

http://www.elasticsearch.org/guide/reference/query-dsl/regexp-query/


karmi:
Thanks — would it be possible to add a simple integration test for the feature?

PR #760, opened 2013-06-19
Add nodes method to Tire::Index to allow retrieval of plugin information

plcstevens:
This adds a plugin method to Tire::Index

We use this method to check the list of installed plugins

```
index = Tire::Index.new('index_name')
index.nodes(plugin: true)
```

NOTE: we could not get all the Percolator tests to pass locally however as we have not modified this code we believe the issue is unrelated

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -119,6 +119,16 @@ module Tire
       MultiJson.decode(@response.body)[@name]['settings']
     end
 
+    def nodes(options={})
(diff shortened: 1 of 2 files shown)


karmi:
Hi, thanks for the contribution, but at this moment, I won't like to extend the Tire features further. I'll gladly merge the patch into https://github.com/karmi/tire-contrib if you extract it and send a pull request.

PR #761, opened 2013-06-19
Don't require Rails 3

ghost:


ghost:
Massive fail.  This wasn't supposed to be here. :(

PR #766, opened 2013-06-21
Prevent load options from being passed to ElasticSearch

jcf:
WebMock can't [parse](https://github.com/bblimke/webmock/blob/master/lib/webmock/util/query_mapper.rb#L96) the load/include params that get sent through to ElasticSearch. Rather than fixing WebMock I've opted for the simpler option of removing the load options from the generated search URL.

**N.B.** The tire build was [failing](https://travis-ci.org/karmi/tire) when I cloned the repo.

I'm going to install Tire from this branch and test whether this resolves the issue with WebMock specifically.

### DIFF
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -48,7 +48,7 @@ module Tire
       end
 
       def params
-        options = @options.except(:wrapper, :payload)
+        options = @options.except(:wrapper, :payload, :load)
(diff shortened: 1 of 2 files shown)


karmi:
@jcf Good catch!, merged in. There are long standing issues with percolator tests, which I haven't yet solved, and some other tests fail on ES master, which I also haven't yet solved.

Issue #762, opened 2013-06-19
Proposal / potential PR: Support for custom load parameters


fgrehm:
Hey @karmi,

I've just started experimenting with the code below and I was wondering if it is something that could be added to Tire's core or to `tire-contrib`:

``` ruby
Tire::Results::Collection.class_eval do
  def __find_records_by_ids(klass, ids)
    if @options[:load].is_a?(Hash)
      klass.find(ids, @options[:load][klass])
    else
      @options[:load] === true ? klass.find(ids) : klass.find(ids, @options[:load])
    end
  end
end
```

The idea would be to support fine grained eager loading on multi model searches. If you think it is a nice idea, I'd be more than happy to turn this issue into a proper pull request either here or on  tire-contrib :)

karmi:
@fgrehm Sounds like a nice idea, but could you please add a more fleshed out example?

fgrehm:
@karmi sure, I'll get back to u on monday :)

fgrehm:
@karmi here's a simple [gist](https://gist.github.com/fgrehm/8b00c564d19433c8ef25) with example models and usage, this is basically what I'm currently doing on the app I'm working on. I hope they are self explanatory but if u need more info just LMK :) 
our current need is to perform a global / multi model search and render associations data on the search results page alongside each result item so we ended up doing that to avoid N+1 queries

karmi:
Will have a look. At the moment, the biggest problem is related to #767, and with the current release of Rails 4, the problem is more imminent.

fgrehm:
@karmi got it, but the issues are not mutually exclusive as there's currently no way to specify custom eager loading behavior per class on multi model searches :) I've [updated the gist](https://gist.github.com/fgrehm/8b00c564d19433c8ef25#file-gistfile1-rb-L13-L16) adding an example of how it could be used with lambdas

PR #767, opened 2013-06-21
Update deprecated ActiveRecord includes syntax

jcf:


karmi:
Good, but I think we need to tread more carefully here. It's bad the former syntax was deprecated, since it allowed not only includes, but also conditions, etc. I'm afraid we can't just change it like this, and we definitely unit and integration tests for this.

jcf:
The chainable API is the preferred way of doing conditions etc. now and has been [since Rails 3.0](http://weblog.rubyonrails.org/2010/8/29/rails-3-0-it-s-done/).

What's your stance on Rails version support in Tire? I think this change will cover Rails 3 & 4, and can look into the tests sometime early next week if you want something like this mergeable into upstream.

karmi:
@jcf I know the hash interface has been deprecated. I haven't really put much thought how to support the chainable interface in the hash options to `search`. Maybe the only use case is `include()`, maybe there are other use cases...

[Old doc: http://apidock.com/rails/v2.3.8/ActiveRecord/Base/find/class]

jcf:
I remember the good old…

``` ruby
User.find(:all, :conditions => {:name => 'Karel'}, :includes => [:elastic_search])
```

…API. If only I could count the number of hashes I must have initialised during my time with Rails 2. Fond memories. :smile:

The first thing that comes to mind is taking a loader block that Tire could use, perhaps yielding the list of IDs to the block so the user can implement find in its entirety.

``` ruby
loader = ->(ids) { Article.includes(:comments).where(published: true).find(ids) }
Article.search(loader: loader) do |s|
  s.query # etc.
end
```

Or perhaps add a method to the class to do the finding with a list of IDs?

``` ruby
class Article
  def self.find_for_tire(ids)
    includes(:comments).where(id: ids)
  end
end
```

I'm sure there's a better way but that's what popped into my mind.

**P.S.** I've force pushed up to my includes branch as I was using `@options[:load]` instead of `@options[:load][:include]`.

Thanks for sharing Tire with the Ruby community. Have a great weekend!

karmi:
Yeah, the good old days :)

I like the idea with `lambda`, since that could allow to leave the old option supported, and evaluate the lambda (polymorphism FTW) as the "finder", exactly as you propose:

``` ruby
Article.search 'love', load: { include: "comments" }
```

``` ruby
Article.search 'love', load: lambda { |ids| includes("comments").find(ids) }
```

Also see #702 and #602 where we discuss some strategies how to mitigate the issue of "record no longer in db".

Also related: #762.

PR #785, opened 2013-07-05
Allow import(query) to take longer than 10 min.

threez:
Snip from the mongodb doc:

By default, the server will automatically close the cursor after 10 minutes of inactivity or if client has exhausted the cursor. To override this behavior, you can specify the noTimeout wire protocol flag in your query; however, you should either close the cursor manually or exhaust the cursor.

http://docs.mongodb.org/manual/core/read-operations/

This especially happens if a lot of data is going to be index.

### DIFF
--- a/lib/tire/model/import.rb
+++ b/lib/tire/model/import.rb
@@ -67,7 +67,7 @@ module Tire
           include Base
           def import &block
             items = []
-            klass.all.each do |item|
+            klass.all.no_timeout.each do |item|
               items << item
               if items.length % options[:per_page] == 0
                 index.import items, options, &block


karmi:
@threez Sounds useful, can we close the cursor after the block then? I saw something `timeout: false` notation, could it be more descriptive then the `no_timeout` method? I have no production familiarity with Mongoid?

threez:
@karmi I prefer the no_timeout method since it is not silently failing like a with(options) block would do. Some of the options might be called different in the future. The specs might not detect the problem because they don't exceed the timeout. So i would stick to the no_timeout method. The no_timeout is only activated for that particular query, so  im not sure what you mean with closing the cursor after the block. Do you mean to kill the cursor if an error occurs?

karmi:
@threez Agreed on `no_timeout`, if you think it's better. I was assuming that you have to manually close the cursor after the operation (based on brief reading of docs), but errors are different thing -- maybe close manually in `ensure` clause?

threez:
Actually i'm not sure, i raised a question to the moped project. Because i think this should be handled in the driver. The driver should not allow leaking of cursors, even if they don't have a timeout. https://github.com/mongoid/moped/issues/206

karmi:
@threez Can you get in touch with @brightbits  from #870, and debug the issue on Mongoid?

PR #789, opened 2013-07-09
Integration test fixes

phoet:
there are a couple of changes that make `rake test:integration` run locally without failing tests. hopefully that also fixes the tests on travis.

### DIFF
--- a/test/integration/active_record_searchable_test.rb
+++ b/test/integration/active_record_searchable_test.rb
@@ -610,10 +610,10 @@ module Tire
 
       end
 
-      context "percolated search" do
+      travis_context "percolated search" do
         setup do
           delete_registered_queries
-          delete_percolator_index if ENV['TRAVIS']
+          delete_percolator_index
           ActiveRecordModelWithPercolation.index.register_percolator_query('alert') { string 'warning' }
           Tire.index('_percolator').refresh
         end
--- a/test/integration/custom_score_queries_test.rb
+++ b/test/integration/custom_score_queries_test.rb
@@ -68,7 +68,7 @@ module Tire
           query do
             # Replace documents score with parameterized computation
             #
-            custom_score :script => "doc['words'].doubleValue / max(a, b)",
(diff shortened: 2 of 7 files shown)


phoet:
@karmi did you ever find out why the percolate queries fail on travis? is that a problem of versions? 

the changes i made allow the local test-suite to pass with ruby 2 and elasticsearch 0.90.2

karmi:
@phoet It works when you wipe _percolator index, but that can't be done normally, people might lose data. Will try to get to all the issues soon, limited online now...

test/integration/active_record_searchable_test.rb
```diff
@@ -613,7 +613,7 @@ module ::Rails; end
       context "percolated search" do
         setup do
           delete_registered_queries
-          delete_percolator_index if ENV['TRAVIS']
+          delete_percolator_index
```
karmi, line 616: @phoet I'd rather have failing tests, then wipe the percolator index for people :) Someone might just run the Tire suite, not realizing we're wiping the index, he might have some hard-to-get-again data there... Let's be rather more careful here. Sorry about the silence about the pull requests, I'm at a vacation now, will try to squeeze some review & merging into next week.

phoet, line 616: hey, no problem!

i think that the tests must not fail locally. is there a way to seperate
these destructive tests somehow and have them run only on travis?

Am Donnerstag, 18. Juli 2013 schrieb Karel Minarik :

karmi, line 616: Yeah, that was my thinking in adding the `if ENV['TRAVIS']`, might be actually better to run percolator tests with this "flag". Obviously, the best solution would be to launch a separate ES service just for tests -- have something on backburner for that :)

karmi:
@phoet Thanks for looking into this! In the end, I've changed the way we run "destructive" tests without introducing the special context, it might be more transparent this way. Hopefully Travis will get green again :)

PR #810, opened 2013-07-26
Update __find_records_by_ids to implement Strategy per ORM.

jeyb:
I haven't done much for the other ORMs as I am not too familiar with them. But this does give a good base to add customizations.


jeyb:
@karmi whatcha think?

karmi:
@jeyb Can you please remove the copy & paste from import code crumbs? :) Also, I think the test could be more readable?

PR #820, opened 2013-07-31
Get rid of Rails 4 deprecation warnings

egze:
Get rid of the DEPRECATION WARNING: Passing options to #find is deprecated. Please build a scope and then call #find on it.


karmi:
This will have to be approached in a more consistent manner, I think; see #767 for discussion.

karmi:
Related #810

PR #824, opened 2013-08-02
Fix #has_attribute? on persistent models.

MitinPavel:
# has_attribute? from Tire::Model::Persistence raises "undefined local variable or method `properties' for ..." error. The pull-request fixes the issue.

By the way, if nobody has ever complained about the issue, it might be more wisdom to just remove the unused method from API.

P. S. Thanks for such good gem.


karmi:
I think at some point in time ActiveModel required/recommended/etc `has_attribute` -- thanks for the fix, I'll have a look into the wider context.

PR #831, opened 2013-08-09
Specify how to connect to a remote ES instance in docs

lreeves:
This is just a quick addition to the README on how to connect to remote Elastic Search indexers (since I didn't see it in the existing examples).

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -51,6 +51,9 @@ To test-drive the core _Elasticsearch_ functionality, let's require the gem:
     require 'tire'
 ```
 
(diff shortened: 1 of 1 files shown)


karmi:
Hi, that's a fair point, but I think we need to mention `Tire.configure { url "xxx" }` as well (see `examples/tire-dsl.rb`)

lreeves:
Geah good point, I'll update the PR tomorrow. Thanks!

lreeves:
@karmi Updated the patch, let me know if that's good.

Nitrodist:
@karmi :shipit:

PR #844, opened 2013-08-17
Don't remove `match` query followed by `ids` query

yaotti:
When the following code is executed, a json (query) containing `ids` and `match` queries should be printed.

``` ruby
Tire.search 'article' do
  query do
    match :foo, 1
    ids [1, 2, 3]
  end
  puts to_json
end
```

However this code prints the following, the query missing `match` block.

``` js
{
  "query": {
    "ids": {
      "values": [
        1,
        2,
        3
      ]
    }
  }
}
```

This PR resolve the problem to build correct queries as follows.

``` js
{
  "query": {
    "bool": {
      "must": [
        {
          "match": {
            "foo": {
              "query": 1
            }
          }
        },
        {
          "ids": {
            "values": [
              1,
              2,
              3
            ]
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/lib/tire/search/query.rb
+++ b/lib/tire/search/query.rb
@@ -92,8 +92,19 @@ module Tire
       end
 
       def ids(values, type=nil)
-        @value = { :ids => { :values => Array(values) }  }
-        @value[:ids].update(:type => type) if type
+        value = { :values => Array(values) }
+        value.update(:type => type) if type
+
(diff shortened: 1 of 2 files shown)


karmi:
Hmm, but this should bw wrapped in a `bool` query, no?

yaotti:
Yes, that's true!
I should have done that :+1:

PR #855, opened 2013-08-31
Add out_of_range? alias for Kaminari support

monfresh:
Kaminari's method is called `out_of_range?`, but tire uses `out_of_bounds?`. I added the alias to make integration with Kaminari easier.


karmi:
Argh, this must have slipped through when the integration has been added. Merging and pushing, thanks!

PR #860, opened 2013-09-05
Cast persistent model attributes according to defined property type on field (when String, Float, or Integer)

FestivalBobcats:
``` ruby
class PersistentArticleWithTypes

  property :title, :type => 'string'
  property :comment_count, :type => 'integer'
  property :average_score, :type => 'float'
end

article = PersistentArticleWithTypes.new :title => 12345, :comment_count => '12', :average_score => 10
assert_equal '12345', article.title
assert_equal 12, article.comment_count
assert_equal 10.0, article.average_score
```

The above example is a model and subsequent unit test showing the desired functionality.  While the example is definitely contrived, this change will prevent extra steps in the controller to cast param values to the appropriate type.  I'm developing a Rails app with Tire as the primary ORM (using ES for storage), and the issue has came up several times now.  Unless I'm totally wrong, I do believe this reflects the default behavior of AR and Mongoid.

### DIFF
--- a/lib/tire/model/persistence/attributes.rb
+++ b/lib/tire/model/persistence/attributes.rb
@@ -121,6 +121,10 @@ module Tire
           # instances and automatically convert UTC formatted strings to Time.
           #
           def __cast_value(name, value)
+
+            property_mapping = self.class.mapping[name.to_sym]
+            type_class = property_mapping[:type].camelize if property_mapping
+
             case
 
               when klass = self.class.property_types[name.to_sym]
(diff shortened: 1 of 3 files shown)


karmi:
Hi, I understand the motivation for this, but think that it would be better to use the "strongly typed" JSON documents here, ie. feed `10.0` to `average_score`, if it indeed should be a float. I'm a bit wary of performance and maintenance implications of checking the value type for all properties. Issue #383 is actually trying to solve similar problem. Maybe it would be better solution? Could you try it, potentially cleaning up the code in that issue?

PR #864, opened 2013-09-07
Lambda/proc property defaults should be called at record initialization as opposed to time of property definition

FestivalBobcats:
``` ruby
class PersistentArticleWithDefaults

  property :created_at, :default => lambda { Time.now }
end
```

As it is currently in the master branch, lambda-based defaults for properties are evaluated at the time of definition.  This causes all records created to have the same value for "created_at", which does not appear to be the intended behavior (unless I am wrong).


karmi:
This is cool, thanks!, embarrassing oversight!

PR #865, opened 2013-09-08
:class instance of a property value should be able to reference the "embedding" doc (or parent)

FestivalBobcats:
In the style of Mongoid's embedded documents, it would be nice (really nice) for property classes to reference their respective "owner" or "parent" docs (`article.comments.first.article`).  The example below will hopefully illustrate the problem.  The only existing solutions are accessor methods in the root model (e.g. `article.comment_posted_within(comment)`) or storing the needed attributes from the root model (such as :created_at in the test) within each embedded object, which is probably a little overkill.

``` ruby
class Comment
  def initialize(article, attrs)
    @article = article
    @attributes = HashWithIndifferentAccess.new(attrs)
  end

  def posted_within
    posted_at - @article.created_at
  end

  # ... method_missing and whatnot
end

class Article

  property :created_at, :type => 'date'
  property :comments, :class => [Comment]
end

now = Time.now
posted_at = now + 10.minutes
article = Article.new :created_at => now,
                      :comments => [{:body => 'comment body', :posted_at => posted_at}]
assert_equal article.comments.first.posted_within, 10.minutes
```

### DIFF
--- a/lib/tire/model/persistence/attributes.rb
+++ b/lib/tire/model/persistence/attributes.rb
@@ -125,9 +125,9 @@ module Tire
 
               when klass = self.class.property_types[name.to_sym]
                 if klass.is_a?(Array) && value.is_a?(Array)
-                  value.map { |v| klass.first.new(v) }
+                  value.map { |v| klass.first.new(self, v) }
                 else
-                  klass.new(value)
+                  klass.new(self, value)
                 end
 
               when value.is_a?(Hash)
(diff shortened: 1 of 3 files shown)


FestivalBobcats:
And apologies for the barrage of successive pull requests...

karmi:
I understand the motivation, and think this would work, but I'm just a bit wary about changing the instantiation in `persistence/attributes` -- all the models would have to respect this signature. That would also mean breaking incompatibility with existing code.

If we could somehow keep the method signature, I'm all for adding it, ideally "injecting" it from the outside?

Issue #744, opened 2013-06-01
The `tire:import:all` Rake task throws an error: uninitialized constant Addres


congshuanglong:
I create a model called Address.When I execute a rake task called tire:import:all,an error, uninitialized constant Addres, is throwed.
I notice some code in the source code file,the path of which is lib/tire/task.rb, should be modified to fix the above problem.
the former:
  klass          = model_filename.classify.constantize
updated to:
  klass          = model_filename.camelize.constantize

karmi:
Thanks for the fix!, added and pushed. (See also #733)

felixbuenemann:
@karmi unfortunately that didn't fix it for me, I had to modify the rake task to depend on the environment and remove `require path` to get it to work, otherwise it failed to load the models, because one model depended on code from app/uploaders.

karmi:
@felixbuenemann So `rake environment tire:import:all` fails for you but when you manually change the task to depend on the environment task, it works? Also, `rake environment` should definitely load all files registered in `config.load_paths`?

felixbuenemann:
@karmi `rake environment tire:import:all` works if I comment `require path`inside task.rb, because that line causes the models to be loaded twice causing issues with overwriting scopes, redefining constants and in this case raising an exception on an embedded enum class.

The same happens if I enter the rails console and do a `require "./app/models/somemodel"`.

Maybe this could be fixed by trying to avoid double require like this?

``` ruby
begin
  klass = model_filename.camelize.constantize
rescue NameError
  require(path) ? retry : raise
end
```

If you think this is a viable solution I could make a PR.

PR #869, opened 2013-09-18
Fix possible double require of classes

felixbuenemann:
When loading environment before tire:import:all, via `rake environment tire:import:all` it is possible that models are already loaded when tire tries to load them.
This can cause redefinition of constants and all other kinds of havoc.

This fix first tries to constantize the models to see if they are already loaded before trying to require them.

See issue #744 for discussion.


felixbuenemann:
@karmi ping

karmi:
I like the solution!, thanks, merged

PR #877, opened 2013-09-23
Skipping model/concerns folder

bitgangsta:
In rails models can have "concerns" (a default pattern in rails4) - which help skinny out models and separate common code among them. But if you try to load them you'll get exceptions since they aren't always prefixed with `concerns::` but yet live in the `model/concerns` folder. This causes the call to `constantize` to fail.

There might be a more elegant way to solve this, but here we have a quick and dirty solution.

What do you think?

### DIFF
--- a/lib/tire/tasks.rb
+++ b/lib/tire/tasks.rb
@@ -124,6 +124,9 @@ namespace :tire do
         require path
 
         model_filename = path[/#{Regexp.escape(dir.to_s)}\/([^\.]+).rb/, 1]
+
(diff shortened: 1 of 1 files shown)


karmi:
Hi, this is quite fine. Glad to see somebody using concerns to separate Tire-related logic. Have you seen articles by Josh Symmonds? http://joshsymonds.com/blog/2012/10/25/rails-concerns-v-searchable-with-elasticsearch/

bitgangsta:
:+1: it's a great article :) Thanks for merging this in!

PR #884, opened 2013-10-05
Disable mongoid's identity map during importing

dzhlobo:
Iterating over all documents in collection with enabled identity map will fill memory until task will be killed by OS.

From mongoid doc: "When a document is now loaded from the database, it is automatically added to the identity map by its class and id. Subsequent request for that document by its id will not hit the database, but rather pull the document back from the identity map itself".

See http://mongoid.org/en/mongoid/docs/identity_map.html for additional information about identity map.

### DIFF
--- a/lib/tire/model/import.rb
+++ b/lib/tire/model/import.rb
@@ -67,11 +67,13 @@ module Tire
           include Base
           def import &block
             items = []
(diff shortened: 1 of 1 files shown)


karmi:
Bump @karmi

PR #904, opened 2013-11-08
Instance eval with delegation proof of concept.

rafael:
Hi,

I've been using tire and I've been facing the following issue. Sometimes, I would like to use some helper methods within the search blocks, but I always get an undefined method exception. In the current state,  the only work around that I found for that so far is to write the search methods like this:

``` ruby

def execute
  Model.tire.search(page: page, load: true) do |s|
    s.query do |q|
      q.boolean do |b|
        b.must { |condition| condition.string "..."  } if tags
        ...
      end
      ...
    end
    s.sort { by :id, "desc" }
  end
end

def tags
    ...
end
```

I notice that this has to do with the use of instance eval inside the DSL. I found this interesting article where they explained a workaround for that:

http://www.dan-manges.com/blog/ruby-dsls-instance-eval-with-delegation

I started a proof of concept to see if it works with tire. It seems to be working. 

Is there a reason why you guys didn't allow this? I could implement this in all the other places, but I wanted to have your feedback, I don't know if I'm missing something here and there is a reason to not do this. 

Cheers!

### DIFF
--- a/lib/tire/model/search.rb
+++ b/lib/tire/model/search.rb
@@ -73,6 +73,7 @@ module Tire
           end
 
           options   = default_options.update(options)
+          options[:context] = eval "self", block.binding if block_given?
           sort      = Array( options.delete(:order) || options.delete(:sort) )
 
           s = Tire::Search::Search.new(options.delete(:index), options)
--- a/lib/tire/search.rb
+++ b/lib/tire/search.rb
@@ -4,7 +4,8 @@ module Tire
 
     class Search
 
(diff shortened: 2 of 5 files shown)


karmi:
Hi, there's been a discussion about this quite some time ago e.g. here: https://gist.github.com/ddemaree/1408149

You're take is interesting, since it has a minimal footprint, and it's quite understandable. However, since Tire was retired, I don't think adding features to it makes real sense? (I'm working on a ActiveRecord/Rails/etc integration for the new gem)

rafael:
Yeah, I thought that about that too. Thanks for the answer! Is it the new gem for Rails live already?

PR #908, opened 2013-11-12
Update index response

spra85:
`Tire::Model::Persistence::Storage` exposes the response from ElasticSearch in the return value of `update_index`.  This pull request does the same for `Tire::Model::Search`, exposes the index response through an attr_reader - mentioned also in this issue from awhile ago (https://github.com/karmi/retire/issues/251).


karmi:
That makes sense, thanks, I'll look into the test fixes commit (332e005) later.

PR #912, opened 2013-11-25
Fixing typo in the boolean queries section of the tire index page

JamesWatling:


karmi:
hey, unfortunately the HTML file is generated from `tire-dsl.rb` and the Markdown processor apparently can't handle `_published_on_` well. Seems like it was handled better in the last generation: http://karmi.github.io/retire/#section-85

PR #929, opened 2014-01-22
Added http error handling to Index#retrieve.

msonnabaum:
Unlike `bulk`, `retrieve` didn't have a check for 500 response codes, which led to Tire::Model returning an object from find instead of reporting the error.

This change adds the same handling as in `bulk`, but makes an exception for 404 since that is handled by the `h['exists'] == false` later on.


karmi:
Thanks, added a test and merged + pushed.

PR #931, opened 2014-01-29
Wrapper lambda support

fabiotomio:


karmi:
Hi, thanks for the pull request, but apart from critical fixes and additions, I think we shouldn't add features to Tire anymore...

PR #935, opened 2014-02-05
Correct debug log level is 'debug'

ryansch:
The project README suggests setting up debug logging with the following:

``` ruby
 Tire.configure { logger 'elasticsearch.log', :level => 'debug' }
```

This pull fixes the `Index#bulk` debug output to test for the 'debug' log level instead of 'verbose'.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -258,7 +258,7 @@ module Tire
         end
 
       ensure
-        data = Configuration.logger && Configuration.logger.level.to_s == 'verbose' ? payload.join("\n") : '... data omitted ...'
+        data = Configuration.logger && Configuration.logger.level.to_s == 'debug' ? payload.join("\n") : '... data omitted ...'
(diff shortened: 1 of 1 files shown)


karmi:
@ryansch That was deliberate, as far as I remember... Otherwise you'd see lines and lines of curl commands.

On the other hand, I can see how `debug` would print the "data omitted" version, and the `verbose` version would print the JSON docs being sent over the wire...

ryansch:
I suppose the problem is setting the log level to `verbose` means I lose all of the other `debug` logging.

karmi:
Agreed... I don't think we can "fix" Tire in this respect, though... I would keep this bug freezed, to be honest, to not change the behaviour on the re-tired project...

ryansch:
Works for me.  We're already using the new elasticsearch gem and slowly letting it take over our project.  :smile:

karmi:
Very glad to hear that :) Let's let this sleep then -- otehrwise, if it would really help you with something, we can make the change.

PR #937, opened 2014-02-13
ruby good style

dimanlin:
I think it's better


karmi:
Hi, thanks for the care :), though, does it really matter to improve such details on a re-tired gem?

dimanlin:
I am not a strong programmer, but I think it looks neater :)

PR #939, opened 2014-02-21
Fixed Index.retrieve so that it works with ES 1.0.0 and previous versions

jonstokes:
The latest version of ES breaks the Index#retrieve function, and it no longer returns nil as it should. This fixes it.

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -364,7 +364,7 @@ module Tire
       wrapper = options[:wrapper] || Configuration.wrapper
       if wrapper == Hash then h
       else
-        return nil if h['exists'] == false
+        return nil if (h['exists'] || h['found']) == false
         document = h['_source'] || h['fields'] || {}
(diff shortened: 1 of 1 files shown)


karmi:
Thanks! Make sure to have a look at the https://github.com/elasticsearch/elasticsearch-rails/tree/master/elasticsearch-model gem.

lib/tire/index.rb
```diff
@@ -364,7 +364,7 @@ def retrieve(type, id, options={})
       wrapper = options[:wrapper] || Configuration.wrapper
       if wrapper == Hash then h
       else
-        return nil if h['exists'] == false
+        return nil if (h['exists'] || h['found']) == false
```
karmi, line 367: @jonstokes This wasn't actually working I think :)

``` ruby
(false || nil) == false
# => false
```

jonstokes, line 367: It works for me.

PR #940, opened 2014-02-22
rename document_type to tire_document_type

flyerhzm:
we have an association named `document_type` which conflicts in retire naming module, so replace `document_type` to `retire_document_type`

### DIFF
--- a/examples/tire-dsl.rb
+++ b/examples/tire-dsl.rb
@@ -88,7 +88,7 @@ Tire.index 'articles' do
       @attributes.each_pair { |name,value| instance_variable_set :"@#{name}", value }
     end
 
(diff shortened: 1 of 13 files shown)


karmi:
Hi, interesting. First, I'm afraid I can rule out renaming `document_type` to `retire_document_type`. The gem has been indeed retired, no features are added, and this would be a 100% breaking compatibility change.

Second, no method defined by Tire should conflict with a method defined by the user -- that's the reason why everything is route through the `tire` proxy class and instance methods. Maybe the `document_type` method has some bug in this respect?

karmi:
So, again -- the `tire` proxy should defend you against a cross-polination here. Closing this, waited for feedback.

PR #944, opened 2014-02-25
Edge case for respond_to? when serializing items

fabn:
I found an issue with class `Tire::Result::Item` when the object is serialized. The current version cannot be deserialized by (for instance) Psych, since it uses [`Class#allocate`](https://github.com/tenderlove/psych/blob/master/lib/psych/visitors/to_ruby.rb#L47) instead of 'Class#new`.

In that case `#initialize` is not called and `@attributes` is nil.

Hope this will get merged.


karmi:
Thanks!, merged the respond_to? fix.

PR #947, opened 2014-03-13
Cast field data that is wrapped in an array from elasticsearch v1

esmarkowski:
Elasticsearch v1 returns field values as an array when using `fields` option.

https://github.com/elasticsearch/elasticsearch/issues/4542

### DIFF
--- a/lib/tire/model/persistence/attributes.rb
+++ b/lib/tire/model/persistence/attributes.rb
@@ -129,6 +129,9 @@ module Tire
           # instances and automatically convert UTC formatted strings to Time.
           #
           def __cast_value(name, value)
+            klass = self.class.property_types[name.to_sym]
+            # elasticsearch v1 returns field data wrapped in an array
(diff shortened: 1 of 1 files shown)


karmi:
Hi Spencer, I guess it makes sense to make this change (although the change in Elasticsearch is correct). So, what about:
1. Squashing the commits into a single one,
2. Adding a unit test for the behaviour.

lib/tire/model/persistence/attributes.rb
```diff
@@ -129,6 +129,9 @@ def __update_attributes(attributes)
           # instances and automatically convert UTC formatted strings to Time.
           #
           def __cast_value(name, value)
+            default = self.class.property_defaults[name.to_sym]
+            # elasticsearch v1 returns field data wrapped in an array
```
karmi, line 133: Should be `Elasticsearch 1.x`

karmi, line 133: Maybe we should add the link to https://github.com/elasticsearch/elasticsearch/issues/4542 as well?

PR #959, opened 2014-04-11
Improve logger & general string concat performance

### LINKED ISSUE #155: When search response is extremely large logging takes a long time
This is an edge case which happens when the query response is extremely large due to large document content and the logger level is in debug mode.

Essentially the logger rendering the response content into separate lines takes long enough that it times out the request if originating from a Rails controller.

Tracked down to line 50 in logger.rb, 
    json.to_s.each_line { |line| content += "# #{line}" } unless json.to_s !~ /\S/

The call to .each_line takes a non-trivial amount of time when the response has hundreds of lines.

Suggested trivial patch:

```
index d8c0122..a84f6cc 100644
--- a/lib/tire/logger.rb
+++ b/lib/tire/logger.rb
@@ -47,7 +47,8 @@ module Tire
       content += " [#{status}]"
       content += " (#{took} msec)" if took
       content += "\n#\n" unless json.to_s !~ /\S/
-      json.to_s.each_line { |line| content += "# #{line}" } unless json.to_s !~ /\S/
+      line_no = 0
+      json.to_s.each_line { |line| content += "# #{line}"; line_no = line_no + 1; break if (line_no > 50) } unless json.to_s !~ /\S/
       content += "\n\n"
       write content
     end
```

bbonamin:
Dramatically improves the performance of the code where strings are concatenated, by changing the "+=" for "<<". 
Sample profiling of [line 50 on logger.rb](https://github.com/karmi/retire/blob/0512cc5e3528d9ab4c709f58a3ffb2e20c1f4196/lib/tire/logger.rb#L50) with a 20,000 lines json response from ElasticSearch:

```
Before the change
99.74%  48.04 seconds String#each_line

After the change
74.91%  0.11 seconds String#each_line
```

Fixes #155

### DIFF
--- a/lib/tire/index.rb
+++ b/lib/tire/index.rb
@@ -93,7 +93,7 @@ module Tire
       end
 
       url  = "#{self.url}/#{type}/_mapping"
-      url += "?#{params.to_param}" unless params.empty?
+      url << "?#{params.to_param}" unless params.empty?
 
       payload = { type => mapping }.to_json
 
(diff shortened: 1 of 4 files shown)


karmi:
Thanks, verified and merged!

PR #983, opened 2015-01-20
fix when alias is missing in value result

diclophis:
@cantino this is what is needed to fix the gem for es1.x as far as I can tell, this should be ok to get into master of this gem which will ease migrations


diclophis:
Oops, this is based from the wrong thing

